PR #98, opened 2017-10-02
Fix SSL issues pointed at #44

### LINKED ISSUE #44: Enable HTTPS protocol

ffdarkpenguin:
closes #44 
- Using port variable to define flask port
- Added pyOpenSSL to requirements.txt
- Moved the code from run.py to app.py
- Moved run.py to run.sh: now it only calls: python oneview-redfish-toolkit/app.py
- Removed prints and now use log system


oneview_redfish_toolkit/app.py
```diff
@@ -62,3 +62,42 @@
 app.register_blueprint(manager)
 app.register_blueprint(odata)
 app.register_blueprint(thermal)
+
+if __name__ == "__main__":
```
ricardoas: We don't need this if for main anymore...

oneview_redfish_toolkit/app.py
```diff
@@ -62,3 +62,42 @@
...
+    # Checking port range
+    if port < 1 or port > 65536:
+        logging.error("Port must be an integer number between 1 and 65536")
+        exit(1)
+
+    ssl_type = config["ssl"]["SSLType"]
+    # Check SSLType:
+    if ssl_type not in ('disabled', 'adhoc', 'certs'):
+        logging.error(
+            "Invalid SSL type: {}. Must be one of: None, adhoc or certs".
```
danielreed: Instead of None here it should say disabled.

oneview_redfish_toolkit/app.py
ffdarkpenguin: source of this please. Don't need also don't make it wrong, is a pattern long used.

oneview_redfish_toolkit/app.py
ffdarkpenguin: fixed

oneview_redfish_toolkit/app.py
ffdarkpenguin: done

PR #101, opened 2017-10-02
add description to manager components

bsoutham:
add description to blade manager ("HPE Integrated Lights Out") and to enclosure manager ( "HPE Onboard Administrator")
modify tests to pass

### DIFF
--- a/oneview_redfish_toolkit/api/blade_manager.py
+++ b/oneview_redfish_toolkit/api/blade_manager.py
@@ -39,9 +39,9 @@ class BladeManager(Manager):
 
         super().__init__(server_hardware, server_hardware['mpFirmwareVersion'])
 
-        self.redfish["FirmwareVersion"] = server_hardware['mpFirmwareVersion']
         self.redfish["ManagerType"] = "BMC"
         self.redfish["Name"] = "Blade Manager"
+        self.redfish["Description"] = "HPE Integrated Lights Out"
         self.redfish["Links"] = collections.OrderedDict()
         self.redfish["Links"]["ManagerForServers"] = list()
         self.redfish["Links"]["ManagerForServers"].append(
--- a/oneview_redfish_toolkit/api/blade_manager.py
+++ b/oneview_redfish_toolkit/api/blade_manager.py
@@ -41,7 +41,7 @@ class BladeManager(Manager):
 
         self.redfish["ManagerType"] = "BMC"
         self.redfish["Name"] = "Blade Manager"
-        self.redfish["Description"] = "HPE Integrated Lights Out"
+        self.redfish["Description"] = "HPE OneView Blade Manager"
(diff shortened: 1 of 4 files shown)


oneview_redfish_toolkit/api/blade_manager.py
```diff
@@ -39,9 +39,9 @@ def __init__(self, server_hardware):
 
         super().__init__(server_hardware, server_hardware['mpFirmwareVersion'])
 
-        self.redfish["FirmwareVersion"] = server_hardware['mpFirmwareVersion']
```
danielreed: Why was this removed?

oneview_redfish_toolkit/api/blade_manager.py
```diff
@@ -39,9 +39,9 @@ def __init__(self, server_hardware):
...
         self.redfish["ManagerType"] = "BMC"
         self.redfish["Name"] = "Blade Manager"
+        self.redfish["Description"] = "HPE Integrated Lights Out"
```
danielreed: This isn't really the iLO that we are exposing in this service.  Shouldn't the description be something more like 'Synergy Blade Manager' or 'HPE OneView Blade Manager'?

oneview_redfish_toolkit/api/enclosure_manager.py
```diff
@@ -41,6 +41,7 @@ def __init__(self, enclosure, oneview_version):
 
         self.redfish["ManagerType"] = "EnclosureManager"
         self.redfish["Name"] = "Enclosure Manager"
+        self.redfish["Description"] = "HPE Onboard Administrator"
```
danielreed: This isn't really the OA that we are exposing in this service. Shouldn't the description be something more like 'Synergy Manager' or 'HPE OneView Manager'?

oneview_redfish_toolkit/api/blade_manager.py
bsoutham: it is passed into the super class and stored.  This is a duplicate and should have been removed.

oneview_redfish_toolkit/api/blade_manager.py
bsoutham: If the manager type is a BMC (baseboard managmenet controller) -- then for HPE gear, that iLO.  Other vendors have other names (BMC, DRAC, etc.)

https://hewlettpackard.github.io/ilo-rest-api-docs/ilo5/?python#find-the-ilo-5-management-processor
seems to indicate that iLO uses them interchangeably.


Are you saying that the description is for the logical redfish manager, and not the underlying hardware component?

oneview_redfish_toolkit/api/enclosure_manager.py
bsoutham: same comment as above

PR #111, opened 2017-10-05
Fix status for System, Chassis and Manager

### LINKED ISSUE #106: Fix status for System, Chassis and Manager
As oneview-redfish-toolkit user, I access any of the three resources using its URI and received a redfish compliant status information composed of state and health, mapping possible OneView values to matching equivalents.

llvieira:
Closes #106.

### DIFF
--- a/oneview_redfish_toolkit/api/status_mapping.py
+++ b/oneview_redfish_toolkit/api/status_mapping.py
@@ -21,7 +21,7 @@ STATUS_MAP = {
     },
     "Disabled": {
         "State": "Disabled",
-        "Health": "OK"
+        "Health": "Warning"
     },
     "Warning": {
         "State": "Enabled",
@@ -33,7 +33,7 @@ STATUS_MAP = {
     },
     "Unknown": {
         "State": "Absent",
-        "Health": "OK"
+        "Health": "Critical"
     }
 }
 
--- a/oneview_redfish_toolkit/api/status_mapping.py
+++ b/oneview_redfish_toolkit/api/status_mapping.py
@@ -33,7 +33,7 @@ STATUS_MAP = {
     },
     "Unknown": {
         "State": "Absent",
-        "Health": "Critical"
+        "Health": None
     }
 }
 
--- a/oneview_redfish_toolkit/api/status_mapping.py
+++ b/oneview_redfish_toolkit/api/status_mapping.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (2017) Hewlett Packard Enterprise Development LP
+#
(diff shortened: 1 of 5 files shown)


oneview_redfish_toolkit/api/status_mapping.py
```diff
@@ -0,0 +1,50 @@
...
+# under the License.
+
+STATUS_MAP = {
+    "OK": {
+        "State": "Enabled",
+        "Health": "OK"
+    },
+    "Disabled": {
+        "State": "Disabled",
+        "Health": "OK"
```
danielreed: Is OK the right status here for Disabled? Should it be Warning or Critical?

oneview_redfish_toolkit/api/status_mapping.py
```diff
@@ -0,0 +1,50 @@
...
+        "State": "Enabled",
+        "Health": "Warning"
+    },
+    "Critical": {
+        "State": "Enabled",
+        "Health": "Critical"
+    },
+    "Unknown": {
+        "State": "Absent",
+        "Health": "OK"
```
danielreed: If its Absent I don't see how it could have an OK status.

oneview_redfish_toolkit/api/status_mapping.py
llvieira: Fixed.

oneview_redfish_toolkit/api/status_mapping.py
llvieira: Fixed.

oneview_redfish_toolkit/api/status_mapping.py
```diff
@@ -0,0 +1,50 @@
...
+        "State": "Enabled",
+        "Health": "Warning"
+    },
+    "Critical": {
+        "State": "Enabled",
+        "Health": "Critical"
+    },
+    "Unknown": {
+        "State": "Absent",
+        "Health": "Critical"
```
danielreed: Question: should Health be Critical or Unknown?  Probably both are okay.

llvieira: In Redfish case is **Critical** or **Warning**, not **Unknown**. I think Critical looks good.

ricardoas: Health can be OK, Warning, Critical or null according to the [spec](http://redfish.dmtf.org/schemas/v1/Resource.json#/definitions/Status)! I was wondering if this is a null case... wdyt?

danielreed: Maybe null is a good fit for this case.

PR #129, opened 2017-10-13
Add device function collection#125

### LINKED ISSUE #125: Implement NetworkDeviceFunctionCollection
As oneview-redfish-user, I access "/redfish/v1/Systems/<server_hardware_uuid>/NetworkAdapters/<deviceId>/NetworkDeviceFunctions" and receive the collection of NetworkDeviceFunctions objects associated with this adapter, one for each 3-tuple (physical_port, virtual_port, port_function). The ID should be the concatenation of each element id separated by "_".

ffdarkpenguin:
closes #125

### DIFF
--- a/oneview_redfish_toolkit/api/network_device_function_collection.py
+++ b/oneview_redfish_toolkit/api/network_device_function_collection.py
@@ -1,0 +1,67 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (2017) Hewlett Packard Enterprise Development LP
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+from oneview_redfish_toolkit.api.redfish_json_validator \
+    import RedfishJsonValidator
+
+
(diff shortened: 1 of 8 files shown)


oneview_redfish_toolkit/api/network_device_function_collection.py
```diff
@@ -0,0 +1,67 @@
...
+        self.redfish["@odata.type"] = \
+            "#NetworkDeviceFunctionCollection.NetworkDeviceFunctionCollection"
+        self.redfish["Name"] = "Network Device Function Collection"
+        members_count = 0
+        self.redfish["Members"] = list()
+        index = int(device_id) - 1
+        devices = server_hardware["portMap"]["deviceSlots"][index]
+        for port in devices["physicalPorts"]:
+            physical_port = str(port["portNumber"])
+            for virtual_port in port["virtualPorts"]:
```
danielreed: You might check with Ricardo but I can't remember if we were going to look at virtual ports.

redfish.conf
```diff
@@ -13,7 +13,7 @@ userName = administrator
 password =
 
 [ssl]
-SSLType = adhoc
+SSLType = disabled
```
danielreed: Did you mean to change this? I prefer disabled.

ricardoas: I prefer too... but not on this patch! It's not NetworkFunction related.

oneview_redfish_toolkit/api/network_device_function_collection.py
ricardoas: @danielreed I though our final agreement was that the NetworkDeviceFunction was a one to one mapping with each virtual port... is it something else?

redfish.conf
ffdarkpenguin: Yes I meant to. It's really a pain have to accept the cert every time during development. Ok, we change it in another patch

PR #137, opened 2017-10-19
Fix device_id range validation

### LINKED ISSUE #136: Fix device_id range validation on network_interface blueprint
current:;  device_id_validation < 0 ....
correct: device_id_validation - 1 < 0 ....

llvieira:
Closes #136.


llvieira:
Ok, I will fix it. It happens because I had not understood so well your description in the issue #136.

PR #139, opened 2017-10-19
Fix issues with slash at the end of the URLs

### LINKED ISSUE #138: Fix issues with slash at the end of the URLs on blueprints
The RESTful operations should work with and without slash at the end of the URLs.

llvieira:
Closes #138.

@ricardoas: with `app.url_map.strict_slashes = False` we no longer have problems with the slash at the end of the route. But I changed the routes to follow a pattern with the slash.


llvieira:
Ok @ffdarkpenguin, let's wait the PO decision.

llvieira:
I will close this PR for now.

PR #154, opened 2017-11-01
Perform schema validation resolving $ref locally

### LINKED ISSUE #135: Perform schema validation resolving $ref locally
Currently the jsonschema validation lib resolves all $ref to DMTF website and download the schemas from that website to complete the validation. We need to find a way for it to refer to files locally available and not use Internet access to download the files.

llvieira:
With this changes we eliminate dependencies between the internet and validation of schemas. Other improvement is that the response time of tests and API is drastically reduced, because no longer needed download the schemas from DMTF website.

Closes #135.


oneview_redfish_toolkit/api/redfish_json_validator.py
```diff
@@ -57,22 +59,37 @@ def _validate(self):
...
             Returns:
                 None
 
             Exception:
                 raises an exception on validation failure
         """
+        schema_dir = util.config['redfish']['schema_dir']
+        schema_paths = glob.glob(os.getcwd() + '/' + schema_dir + '/*.json')
+
+        store = {}
```
ricardoas: Good! You are building a local store for the RefResolver... not sure if redfish_json_validator is the best place to load this info. This way, you'll end up loading/reloading all schema_dir files every time you instantiate an object.

Would utils be a better place to load this? We could add to globals and just import here...

llvieira: Great, I'll fix it.

PR #174, opened 2017-11-07
Adding ETag in the response header

### LINKED ISSUE #160: Does OneView resources implement eTags?
Check if ServerHardware, Enclosure and Rack return eTag information that can be used to mark redfish equivalent resources. If so, implement this as a header on the responses. It should be on the format: ETag: W/"\<string\>" according to [the specification](http://redfish.dmtf.org/schemas/DSP0266_1.1.html#etags)

adriane-cardozo:
Fixed #160 

Adding ETag in the response header for computer system and chassis.

### DIFF
--- a/oneview_redfish_toolkit/tests/blueprints/test_chassis.py
+++ b/oneview_redfish_toolkit/tests/blueprints/test_chassis.py
@@ -157,31 +157,6 @@ class TestChassis(unittest.TestCase):
         self.assertEqual(status.HTTP_200_OK, response.status_code)
         self.assertEqual("application/json", response.mimetype)
         self.assertEqual(self.enclosure_chassis_mockup, json_str)
-
-    @mock.patch.object(util, 'get_oneview_client')
-    def test_get_enclosure_chassis_with_etag(
-            self, get_oneview_client_mockup):
-        """"Tests EnclosureChassis and ETag response header"""
-
-        oneview_client = get_oneview_client_mockup()
-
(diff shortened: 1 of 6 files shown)


oneview_redfish_toolkit/tests/blueprints/test_chassis.py
```diff
@@ -159,6 +159,34 @@ def test_get_enclosure_chassis(
...
+            "/redfish/v1/Chassis/0000000000A66101"
+        )
+
+        json_str = response.data.decode("utf-8")
+
+        # Tests response
+        self.assertEqual(status.HTTP_200_OK, response.status_code)
+        self.assertEqual("application/json", response.mimetype)
+        self.assertEqual(self.enclosure_chassis_mockup, json_str)
+        self.assertEqual(
```
ffdarkpenguin: I would just add this assert line to the enclosure_chassis test. This work, is OK but the other is optimized in my opinion. Same for the other tests.

ricardoas: I kind of agree with this... as there should not be a enclosure chassis without an etag from now on.
Can you update that, @adriane-cardozo?

adriane-cardozo:
@ffdarkpenguin ok, I have just updated the unit tests, moving ETag test to the existing unit tests.
@ricardoas ok, I have just added ETag for Managers and updated unit tests.

PR #175, opened 2017-11-08
Add https

### LINKED ISSUE #162: Warn user when not using https with valid certificate
The user should see a warning message on logs if starting the service with https disabled or using a self-signed ("adhoc" value) certificate.

### LINKED ISSUE #163: Revert to https by default
The default value for SSLType on redfish.conf should be "certs"

### LINKED ISSUE #164: Generate a self-signed on first run
If the user has not changed `redfish.conf` file informing a valid certificate (or disabling the "certs" mode), the service should generate a self-signed certificate on the first run, named `default-self-signed.crt` and `default-self-signed.key`. This certificate will be generated using configuration values available in `redfish.conf`.

ffdarkpenguin:
closes #162 closes #163 closes #164


oneview_redfish_toolkit/util.py
```diff
@@ -284,3 +286,66 @@ def get_oneview_client():
...
+                .crt to certificate file and .key to key file
+            key_length: key length in bits
+            key_type: cryto type: RSA or DSA; defaults to RSA
+        Returns:
+            Nothing
+        Exceptions:
+            Raise exceptions on error
+    """
+
+    config = globals()['config']
```
ricardoas: What happened here? I thought `import util.config as config` was supposed to work to avoid this kind of thing...

oneview_redfish_toolkit/util.py
```diff
@@ -284,3 +286,66 @@ def get_oneview_client():
...
+    private_key = OpenSSL.crypto.PKey()
+    if key_type == "rsa":
+        private_key.generate_key(OpenSSL.crypto.TYPE_RSA, key_length)
+    else:
```
ricardoas: Should we elif to "dsa" and else to error message for invalid key_type?

oneview_redfish_toolkit/util.py
```diff
@@ -284,3 +286,66 @@ def get_oneview_client():
...
+
+    cert.set_serial_number(1)
+    cert.gmtime_adj_notBefore(0)
+    cert.gmtime_adj_notAfter(10 * 365 * 24 * 60 * 60)
+    cert.set_issuer(cert.get_subject())
+    cert.set_pubkey(private_key)
+    cert.sign(private_key, "sha1")
+
+    # Save Files
+    open(os.path.join(dir_name, file_name + ".crt"), "wt").write(
```
ricardoas: Out of curiosity: does the code below (using with) closes file descriptor? Should we change to it?
```
with open(...) as certificate_file:
   dump content
`` `

File: redfish.conf [hunk 4]
```diff
@@ -14,9 +14,18 @@ userName = administrator
...
 
 [ssl]
-SSLType = adhoc
-SSLCertFile =
-SSLKeyFile =
+SSLType = self-signed
+SSLCertFile = certs/self-signed.crt
+SSLKeyFile = certs/self-signed.key
+
+[ssl-cert-defaults]
```
ricardoas: @danielreed @bsoutham: Does HPE OneView related stuff have default values to self-signed certificates?

oneview_redfish_toolkit/util.py
ffdarkpenguin: import util.config on the util file. This is very strange , never imported a file in the same file and really didn't know i could do that. Does it work? Does it make sense? I look way too strange to my eyes.

oneview_redfish_toolkit/util.py
ffdarkpenguin: Yes I'll fix that.

oneview_redfish_toolkit/util.py
ffdarkpenguin: OK

bsoutham:
default is a sha256RSA cert with a 10 year validity (from date of creation) and a 2048 bit key
subject would be something like:
CN = ov-1.gd.demo
O = Hewlett Packard Enterprise
L = Palo Alto
S = California
C = US

redfish.conf
ffdarkpenguin: @ricardoas Fixed with values informed by Blaine.

ricardoas: thanks!

README.md
```diff
@@ -80,12 +80,28 @@ In order to start up oneview-redfish-toolkit service, there is some mandatory co
 
 * `ssl` section
 
-  * **SSLType**: How to configure SSL. One of: disabled (no SSL), adhoc (SSL is enabled with self-signed keys generated by the server), certs (SSL enabled with keys provided by user in the fields below). The default value used is adhoc.
+  * **SSLType**: How to configure SSL. One of: disabled (no SSL), adhoc (SSL is enabled with self-signed keys generated by the server every time you start the server), self-signed (SSL is enabled with a self-signed cert generated in the certs directory if no files named self-signed.crt and self-signed.key exists in that directory. This will create the certificates on the first run and every time you delete the files and restart the server. The directory **certs must** exist in the system root directory) certs (SSL is enabled with keys provided by user in the fields below). The default value used is adhoc.
```
ricardoas: We should work a bit more on this in a follow up...

PR #183, opened 2017-11-17
Fix install problems

ricardoas:
run.sh was not working outside virtual environment as it did not
declared explicitly the python path to be used
(oneview-redfish-toolkit root folder)

Also, using the launched version, it was noticed that pbr lib was
mistakenly considering this service as a lib that should be
installed using pip thus having a version number associated to it
that could not be loaded at runtime.

We also reverted the default security option to adhoc so that a
new self signed certificate is generated by flask every time the
service is launched instead of disabling by default.


oneview_redfish_toolkit/__init__.py
```diff
@@ -1,19 +0,0 @@
...
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
-# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
-# License for the specific language governing permissions and limitations
-# under the License.
-
-import pbr.version
-
-
-__version__ = pbr.version.VersionInfo(
```
ffdarkpenguin: I think this is responsible for being able to inform the version of our app. I think this should be fixed instead of removed.

run.sh
```diff
@@ -14,5 +14,5 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations
 # under the License.
-
-python oneview_redfish_toolkit/app.py
+export PYTHONPATH=$PYTHONPATH:.
```
ffdarkpenguin: I was looking at activate script it unsets any pythonpath variable. So chances are that 99.9% of time this will be empty as we a enforcing the use virtual env. The resulting path will be :. (NOTHING:.) I think this is an invalid path. I recommend this:
PYTHONPATH=${PYTHONPATH:+$PYTHONPATH:}.
If pythonpath is not empty add it to new pythonpath, otherwise set it only to .

PR #218, opened 2018-02-05
Implement POST Subscription

### LINKED ISSUE #206: Add subscription
As oneview-redfish-toolkit user, I am able to add subscription issuing a POST request to "/redfish/v1/EventService/EventSubscriptions".

llvieira:
Closes #206.

### DIFF
--- a/oneview_redfish_toolkit/api/subscription.py
+++ b/oneview_redfish_toolkit/api/subscription.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (2017) Hewlett Packard Enterprise Development LP
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+from oneview_redfish_toolkit.api.redfish_json_validator \
+    import RedfishJsonValidator
+
+
+class Subscription(RedfishJsonValidator):
+    """Creates the Subscription class
+
+        Populates self.redfish with Subscription response.
+    """
+
+    SCHEMA_NAME = 'EventDestination'
+
+    def __init__(self, subscription_id, event_types):
+        """Session constructor
+
+            Populates self.redfish with Subscription response.
+        """
+
+        super().__init__(self.SCHEMA_NAME)
+
+        self.redfish["@odata.type"] = \
+            "#EventDestination.v1_2_0.EventDestination"
+        self.redfish["Id"] = subscription_id
+        self.redfish["Name"] = "EventSubscription " + subscription_id
+        self.redfish["Destination"] = "http://www.dnsname.com/Destination1"
+        self.redfish["EventTypes"] = event_types
+        self.redfish["Context"] = "WebUser3"
+        self.redfish["Protocol"] = "Redfish"
+        self.redfish["@odata.context"] = \
+            "/redfish/v1/$metadata#EventDestination.EventDestination"
+        self.redfish["@odata.id"] = "/redfish/v1/EventService/Subscriptions/1"
+
+        self._validate()
--- a/oneview_redfish_toolkit/api/subscription.py
+++ b/oneview_redfish_toolkit/api/subscription.py
@@ -26,7 +26,7 @@ class Subscription(RedfishJsonValidator):
 
     SCHEMA_NAME = 'EventDestination'
 
-    def __init__(self, subscription_id, event_types):
+    def __init__(self, subscription_id, destination, event_types, context):
         """Session constructor
 
             Populates self.redfish with Subscription response.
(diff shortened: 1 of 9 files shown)


oneview_redfish_toolkit/blueprints/subscription.py
```diff
@@ -0,0 +1,104 @@
...
+        try:
+            body = request.get_json()
+            event_types = body["EventTypes"]
+        except Exception:
+            raise OneViewRedfishError(
+                {"errorCode": "INVALID_INFORMATION",
+                 "message": "Invalid JSON key. The JSON request body"
+                            " must have the key EventTypes."})
+
+        subscription_id = str(len(util.all_subscriptions) + 1)
```
ricardoas: Why? Can't we use UUID as all the other resources?

oneview_redfish_toolkit/blueprints/subscription.py
```diff
@@ -0,0 +1,104 @@
...
+
+        # Build Subscription object and validates it
+        sc = Subscription(subscription_id, event_types)
```
ricardoas: I think there's more to come from the user when subscribing... where is the callback URL? From the json schema, I see the mandatory fields when creating a subscriptions are: 
```
requiredOnCreate: [
"Destination",
"EventTypes",
"Protocol"
],
```

oneview_redfish_toolkit/blueprints/subscription.py
```diff
@@ -0,0 +1,104 @@
...
+
+        for event_type in event_types:
+            if event_type not in util.subscriptions_by_type:
```
ricardoas: If you use UUID, it will never be... if you use a sequential id, then you will have problems when users start deleting subscriptions, wdyt?

oneview_redfish_toolkit/api/subscription.py
```diff
@@ -0,0 +1,49 @@
...
+
+        super().__init__(self.SCHEMA_NAME)
+
+        self.redfish["@odata.type"] = \
+            "#EventDestination.v1_2_0.EventDestination"
+        self.redfish["Id"] = subscription_id
+        self.redfish["Name"] = "EventSubscription " + subscription_id
+        self.redfish["Destination"] = "http://www.dnsname.com/Destination1"
+        self.redfish["EventTypes"] = event_types
+        self.redfish["Context"] = "WebUser3"
```
ricardoas: Where is this value coming from? What does it mean?

oneview_redfish_toolkit/api/subscription.py
```diff
@@ -0,0 +1,49 @@
...
+            Populates self.redfish with Subscription response.
+        """
+
+        super().__init__(self.SCHEMA_NAME)
+
+        self.redfish["@odata.type"] = \
+            "#EventDestination.v1_2_0.EventDestination"
+        self.redfish["Id"] = subscription_id
+        self.redfish["Name"] = "EventSubscription " + subscription_id
+        self.redfish["Destination"] = "http://www.dnsname.com/Destination1"
```
ricardoas: This should come from the user...

oneview_redfish_toolkit/api/subscription.py
```diff
@@ -0,0 +1,49 @@
...
+    def __init__(self, subscription_id, event_types):
+        """Session constructor
+
+            Populates self.redfish with Subscription response.
+        """
+
+        super().__init__(self.SCHEMA_NAME)
+
+        self.redfish["@odata.type"] = \
+            "#EventDestination.v1_2_0.EventDestination"
```
ricardoas: Should this version map directly to the one loaded from schema folder?

oneview_redfish_toolkit/api/subscription.py
llvieira: The field `Context` is required by the schema. The description that I found about that says: *This property shall contain a client supplied context that will remain with the connection through the connections lifetime.*

So, I think it should come from the user too.

oneview_redfish_toolkit/blueprints/subscription.py
llvieira: The required value is an `Id`, that according the spec it is different than `UUID`. Should we generate a random `UUID`?

oneview_redfish_toolkit/api/subscription.py
llvieira: Sure.

oneview_redfish_toolkit/blueprints/subscription.py
llvieira: By the schema, the `Protocol` must have the value `Redfish`. So, it can be hard-coded.

`"Protocol": {
                    "$ref": "#/definitions/EventDestinationProtocol", ...}`

`
EventDestinationProtocol": {
            "type": "string",
            "enum": [
                "Redfish"
            ]
}
`

oneview_redfish_toolkit/blueprints/subscription.py
llvieira: The required value is an `Id`, that according the spec it is different than `UUID`. Should we generate a random `UUID`?

oneview_redfish_toolkit/api/subscription.py
llvieira: Are we going to refactor this in all resource classes? It sounds good, since if the schema version is changed the `@odata.type` will have a wrong information.

ricardoas: ok, moving this to a separate issue.

oneview_redfish_toolkit/blueprints/subscription.py
ricardoas: I know... from the spec the `Id` is a string, so we are allowed to store whatever we want in it, as we've been doing for all resources implemented until this point. If not, then we should think of a new strategy to fill this for ComputerSystem, Chassis, Manager, etc...

If we agree on using UUID, then in my opinion generate a new one for this field is our best and safest approach. Using incremental IDs as it is implemented here will lead to error right after deleting a subscription...

oneview_redfish_toolkit/blueprints/subscription.py
```diff
@@ -0,0 +1,104 @@
...
+                util.subscriptions_by_type[event_type] = dict()
```
ricardoas: This should not be created here... we should've already loaded the dictionaries in `util.py`!

oneview_redfish_toolkit/blueprints/subscription.py
llvieira: Ok. So, I'm using the Python UUID lib to generate it. As I'm using the `uuid1` it has zero chances of collision with a previous generated value, since it takes data/time.

oneview_redfish_toolkit/blueprints/subscription.py
llvieira: We will use UUID.

oneview_redfish_toolkit/blueprints/subscription.py
llvieira: From `util.py` we are loading an empty dictionary, like `{}`. So, what I'm doing here is `{"Alert": {}, ...}`, and after that `{"Alert": {UUID: subscription, ...}, ...}`. I think it is right based on what we defined before.

llvieira: Okay, fixed!

PR #222, opened 2018-02-12
add scmb connection

### LINKED ISSUE #205: Open SCMB connection
As oneview-redfish-toolkit user, when I configure authentication mode to conf file based and if the these credentials have administrator priviledge, I connect to HPE OneView SCMB. If this connection fail, the user should receive an proper error message.

ffdarkpenguin:
closes #205

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,11 @@ In order to start up oneview-redfish-toolkit service, there is some mandatory co
 
 Note: HPE OneView' credentials are stored in clear-text. Make sure only authorized users can access this file.
 
+## SCMB notes:
+
+SCMB event listenning will only be started if authentication_mode is set to `conf`. If necessary the app will request OneView's to generate SCMB certs file and/or download the certs to the correct location. The certs file are: **oneview\_ca**: OneView's CA cert file located at: `certs/oneview_ca.pem`. **scmb\_cert**: OneView's SCMB Client cert file located at: `certs/oneview_scmb.pem`. **scmb\_key**: OneView's SCMB Client key file located at: `certs/oneview_scmb.key`
+
+
 ## Contributing
 
 You know the drill. Fork it, branch it, change it, commit it, and pull-request it. We are passionate about improving this project, and are glad to accept help to make it better. However, keep the following in mind:
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ Note: HPE OneView' credentials are stored in clear-text. Make sure only authoriz
 
 ## SCMB notes:
 
-SCMB event listenning will only be started if authentication_mode is set to `conf`. If necessary the app will request OneView's to generate SCMB certs file and/or download the certs to the correct location. The certs file are: **oneview\_ca**: OneView's CA cert file located at: `certs/oneview_ca.pem`. **scmb\_cert**: OneView's SCMB Client cert file located at: `certs/oneview_scmb.pem`. **scmb\_key**: OneView's SCMB Client key file located at: `certs/oneview_scmb.key`
+SCMB event listening will only be started if authentication_mode is set to `conf`. If necessary the app will request OneView to generate SCMB certs file and/or download the certs to the correct location. The certs file are: **oneview\_ca**: OneView's CA cert file located at: `certs/oneview_ca.pem`. **scmb\_cert**: OneView's SCMB Client cert file located at: `certs/oneview_scmb.pem`. **scmb\_key**: OneView's SCMB Client key file located at: `certs/oneview_scmb.key`
 
 
 ## Contributing
--- a/README.md
+++ b/README.md
@@ -139,9 +139,9 @@ In order to start up oneview-redfish-toolkit service, there is some mandatory co
 
 Note: HPE OneView' credentials are stored in clear-text. Make sure only authorized users can access this file.
 
-## SCMB notes:
+## Event Service notes:
 
-SCMB event listening will only be started if authentication_mode is set to `conf`. If necessary the app will request OneView to generate SCMB certs file and/or download the certs to the correct location. The certs file are: **oneview\_ca**: OneView's CA cert file located at: `certs/oneview_ca.pem`. **scmb\_cert**: OneView's SCMB Client cert file located at: `certs/oneview_scmb.pem`. **scmb\_key**: OneView's SCMB Client key file located at: `certs/oneview_scmb.key`
+Currently, Event Service works only when authentication_mode is set to conf. As it connects directly to HPE OneView SCMB, the toolkit will request OneView to generate SCMB certs and/or download the certs to the correct location. The certs file are: **oneview\_ca**: OneView's CA cert file located at: `certs/oneview_ca.pem`. **scmb\_cert**: OneView's SCMB Client cert file located at: `certs/oneview_scmb.pem`. **scmb\_key**: OneView's SCMB Client key file located at: `certs/oneview_scmb.key`
 
 
 ## Contributing
--- a/oneview_redfish_toolkit/api/scmb.py
+++ b/oneview_redfish_toolkit/api/scmb.py
@@ -100,7 +100,7 @@ def is_cert_working_with_scmb():
         channel.close()
         scmb_conn.close()
     except Exception:
-        logging.exception("Failed to test scmb connenction")
+        logging.exception("Failed to test scmb connection")
         return False
     return True
 
--- a/oneview_redfish_toolkit/api/scmb.py
+++ b/oneview_redfish_toolkit/api/scmb.py
@@ -33,13 +33,7 @@ SCMB_KEY = "certs/oneview_scmb.key"
 
 
 def check_cert_exist():
-    if not os.path.isfile(ONEVIEW_CA):
-        return False
-    if not os.path.isfile(SCMB_CERT):
-        return False
-    if not os.path.isfile(SCMB_KEY):
-        return False
-    return True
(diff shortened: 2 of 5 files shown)


oneview_redfish_toolkit/api/scmb.py
```diff
@@ -0,0 +1,139 @@
...
+                    'server_side': False})
+
+    scmb_connection = pika.BlockingConnection(
+        pika.ConnectionParameters(
+            scmb_server, 5671, credentials=ExternalCredentials(),
+            ssl=True, ssl_options=ssl_options))
+    return scmb_connection
+
+
+def test_cert():
```
danielreed: This function has a weird name. Should this be called from listen_scmb and not called from the app.py?

oneview_redfish_toolkit/app.py
```diff
@@ -233,6 +235,22 @@ def not_implemented(error):
...
 
+    if util.config['redfish']['authentication_mode'] == 'conf':
+        # Loading scmb connection
+        if scmb.check_cert_exist():
+            logging.info('SCMB certs already exists testing connection...')
+        else:
+            logging.info('SCMB certs not found. Generating/getting certs....')
+            scmb.get_cert()
+            logging.info('Got certs. Testing connection...')
+        scmb.test_cert()
```
danielreed: Should this be called here?  It seems like if you have certs then you should connect and start listening for messages.  No need to 'test' the certs.

README.md
```diff
@@ -139,6 +139,11 @@ In order to start up oneview-redfish-toolkit service, there is some mandatory co
 
 Note: HPE OneView' credentials are stored in clear-text. Make sure only authorized users can access this file.
 
+## SCMB notes:
+
+SCMB event listenning will only be started if authentication_mode is set to `conf`. If necessary the app will request OneView's to generate SCMB certs file and/or download the certs to the correct location. The certs file are: **oneview\_ca**: OneView's CA cert file located at: `certs/oneview_ca.pem`. **scmb\_cert**: OneView's SCMB Client cert file located at: `certs/oneview_scmb.pem`. **scmb\_key**: OneView's SCMB Client key file located at: `certs/oneview_scmb.key`
```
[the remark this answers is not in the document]
ffdarkpenguin: fixed

oneview_redfish_toolkit/api/scmb.py
ffdarkpenguin: It's goal is to make sure that the certs I have work with OneView's rabbitMQ. That's why 'test_cert'. It establishes a connection to rabbitMQ and tries to bind to queue. I was not sure that only calling connect would be enough so i decided to go all the way into queue listening to make sure the connection with the cert works fine. The reason I'm calling this from the app is because it's a test. I wish to call listen_scmb (which very soon will be a bunch of threads) when I'm positive that the connection works, I didn't what to do is several times when I'm in a loop to launch threads. If you still find the function name weird now that you know it's purpose  please feel free to to suggest a better name. If you still think that the test location taking in consideration my arguments let me know, I can change that, but still think it's not the best place to put it.

oneview_redfish_toolkit/app.py
ffdarkpenguin: If you are confident that cert tests are unnecessary I can remove the test. I took in consideration that the certs can be modified by user, os the download may be corrupted. If you think it's a extreme care I'll remove the call and the function.

README.md
```diff
@@ -139,6 +139,11 @@ In order to start up oneview-redfish-toolkit service, there is some mandatory co
 
 Note: HPE OneView' credentials are stored in clear-text. Make sure only authorized users can access this file.
 
+## SCMB notes:
+
+SCMB event listening will only be started if authentication_mode is set to `conf`. If necessary the app will request OneView to generate SCMB certs file and/or download the certs to the correct location. The certs file are: **oneview\_ca**: OneView's CA cert file located at: `certs/oneview_ca.pem`. **scmb\_cert**: OneView's SCMB Client cert file located at: `certs/oneview_scmb.pem`. **scmb\_key**: OneView's SCMB Client key file located at: `certs/oneview_scmb.key`
```
ricardoas: is oneview_ca the same used with ssl=certs???

README.md
```diff
@@ -139,6 +139,11 @@ In order to start up oneview-redfish-toolkit service, there is some mandatory co
 
 Note: HPE OneView' credentials are stored in clear-text. Make sure only authorized users can access this file.
 
+## SCMB notes:
```
ricardoas: Event Service notes:

README.md
ricardoas: Currently, Event Service works only when `authentication_mode` is set to `conf`. As it connects directly to HPE OneView SCMB, the toolkit will request OneView to generate...

oneview_redfish_toolkit/api/scmb.py
```diff
@@ -0,0 +1,138 @@
...
+
+from oneview_redfish_toolkit import util
+
+
+ONEVIEW_CA = "certs/oneview_ca.pem"
+SCMB_CERT = "certs/oneview_scmb.pem"
+SCMB_KEY = "certs/oneview_scmb.key"
+
+
+def check_cert_exist():
```
ricardoas: is this the same as `return os.path.isfile(ONEVIEW_CA) & os.path.isfile(SCMB_CERT) & os.path.isfile(SCMB_KEY)`? Will we need a different treatment for each case?

oneview_redfish_toolkit/api/scmb.py
```diff
@@ -0,0 +1,138 @@
...
+            "commonName": "default",
+            "type": "RabbitMqClientCertV2"
+        }
+        util.ov_client.certificate_rabbitmq.generate(cert_info)
+    except HPOneViewException as e:
+        # Cert with that commonName already exists. We are going to get it
+        if e.oneview_response["errorCode"] == 'RABBITMQ_CLIENTCERT_CONFLICT':
+            logging.info('Certs already exists in oneview')
+        else:
+            # Another error is not expected, we raise.
```
ricardoas: should we "loggging.exception" it?

oneview_redfish_toolkit/api/scmb.py
ricardoas: I prefer that only unit/function tests start with `test_`; a better option would be `is_cert_working_with_scmb` based on the usage of it below...

oneview_redfish_toolkit/app.py
```diff
@@ -233,6 +235,24 @@ def not_implemented(error):
...
 
+    if util.config['redfish']['authentication_mode'] == 'conf':
+        # Loading scmb connection
+        if scmb.check_cert_exist():
+            logging.info('SCMB certs already exists testing connection...')
+        else:
+            logging.info('SCMB certs not found. Generating/getting certs....')
+            scmb.get_cert()
+            logging.info('Got certs. Testing connection...')
+        if scmb.test_cert() is False:
```
ricardoas: `if not is_cert_working_with_scmb():`

oneview_redfish_toolkit/app.py
```diff
@@ -233,6 +235,24 @@ def not_implemented(error):
...
+            logging.error('Failed to connect to scmb. Aborting...')
+            exit(1)
+        scmb_thread = Thread(target=scmb.listen_scmb)
+        scmb_thread.daemon = True
+        scmb_thread.start()
+    else:
+        logging.warning("Authentication mode set to session. SCMB events will "
```
ricardoas: Win!

requirements.txt
```diff
@@ -8,4 +8,6 @@ flask-api
 hpOneView
 jsonschema
 pbr>=2.0 # Apache-2.0
+pika
```
ricardoas: so ampq is not working at all?

README.md
ffdarkpenguin: fixed

README.md
ffdarkpenguin: There's no CA in ssl certs.

oneview_redfish_toolkit/api/scmb.py
ffdarkpenguin: fixed

oneview_redfish_toolkit/api/scmb.py
ffdarkpenguin: fixed

oneview_redfish_toolkit/app.py
ffdarkpenguin: ????

requirements.txt
ffdarkpenguin: it is not working with cert auth as oneview don't accept  user auth can't tell you if with other auth mechs it would be ok

oneview_redfish_toolkit/api/scmb.py
ffdarkpenguin: What about the discussion if we should test or not before launching threads?!?!?
The name will be fixed

oneview_redfish_toolkit/app.py
ricardoas: Thanks for warning the user!! :)

oneview_redfish_toolkit/api/scmb.py
ricardoas: I missed this... but I think that we can fail the whole toolkit gracefully the way it is implemented right now. We could wait and get the result of opening the new thread and then fail but it would be more troublesome.

However, @danielreed, if your idea is to leave the toolkit endpoints up even if the SCMB could not be opened... then I need to think a bit more about this scenario!

PR #224, opened 2018-02-14
add event class

### LINKED ISSUE #208: Map OneView events to Redfish
Translate HPE OneView events to Redfish schema according to the mapping identified in #201

ffdarkpenguin:
closes #208


oneview_redfish_toolkit/api/event.py
```diff
@@ -0,0 +1,61 @@
...
+        super().__init__(self.SCHEMA_NAME)
+
+        id = str(os.path.basename(oneview_alert["resourceUri"]))
+
+        self.redfish["@odata.type"] = "#Event.v1_2_0.Event"
+        self.redfish["Id"] = id
+        self.redfish["Name"] = oneview_alert["resource"]["description"]
+        self.redfish["Events"] = []
+        self.redfish["@odata.context"] = "/redfish/v1/$metadata#Event.Event"
+        event_id = 0
```
ricardoas: Should we use the same id we're using on "Id" here?

oneview_redfish_toolkit/api/event.py
```diff
@@ -0,0 +1,61 @@
...
+        event = {}
+        event["EventId"] = str(event_id)
+        event["EventType"] = event_type
+        # Nothing good matches here!!!!!
+        event["MessageId"] = "Base.1.1.Success"
+        event["Severity"] = oneview_alert["resource"]["severity"]
+        event["EventTimestamp"] = oneview_alert["resource"]["created"]
+        self.redfish["Events"].append(event)
```
ricardoas: This is an array of EventRecords, right? We'll need this in a follow up PR. I'd prefer this array to be an empty one and not contain a single empty element now. Is that ok?

oneview_redfish_toolkit/api/event.py
```diff
@@ -0,0 +1,61 @@
...
+class Event(RedfishJsonValidator):
+    """Creates a Event Redfish dict
+
+        Populates self.redfish with some hardcoded Event
+        values and with the response from OneView.
+    """
+
+    SCHEMA_NAME = 'Event'
+
+    def __init__(self, oneview_alert, event_type):
```
ricardoas: There are some info needed in the event that is stored in the subscription object... and we will need to update it on this event before delivering to each client, right?

oneview_redfish_toolkit/mockups/redfish/Alert.json
```diff
@@ -0,0 +1 @@
+{"Name": "Cold Start.", "@odata.context": "/redfish/v1/$metadata#Event.Event", "Events": [{"Severity": "OK", "EventType": "Alert", "EventTimestamp": "2018-02-12T20:12:03.207Z", "EventId": "0", "MessageId": "Base.1.1.Success"}], "Id": "49", "@odata.type": "#Event.v1_2_0.Event"}
```
ricardoas: Can this be indented as all the other ones are?

oneview_redfish_toolkit/api/event.py
ffdarkpenguin: if the array has more than one element what will be the value for the second?!? and the next....?

oneview_redfish_toolkit/api/event.py
ffdarkpenguin: this is not a single empty element, If you called this object you passed and event as a parameter so this will be the info on the event itself.

oneview_redfish_toolkit/api/event.py
ffdarkpenguin: i don't know what you refer to. Please let me know which fields and where to get it.

oneview_redfish_toolkit/api/event.py
ricardoas: oops... I got it!

oneview_redfish_toolkit/api/event.py
ricardoas: "Context" on "EventRecord" is defined as: "A context can be supplied at subscription time. This property is the context value supplied by the subscriber."

Maybe we'll have more from this kind...

PR #227, opened 2018-02-21
Implements DELETE Subscription

### LINKED ISSUE #207: Delete subscription
As oneview-redfish-toolkit user, I am able to delete subscriptions to this event service issuing a DELETE request to "/redfish/v1/EventService/EventSubscriptions/\<id\>" specifying the id of the subscription that should be removed.

llvieira:
Closes #207.

### DIFF
--- a/oneview_redfish_toolkit/blueprints/subscription.py
+++ b/oneview_redfish_toolkit/blueprints/subscription.py
@@ -129,9 +129,6 @@ def remove_subscription(id):
 
         Args:
             id: The Subscription ID.
-
-        Returns:
-            Subscription JSON.
     """
     try:
         sc = util.all_subscriptions[id]
--- a/oneview_redfish_toolkit/blueprints/subscription.py
+++ b/oneview_redfish_toolkit/blueprints/subscription.py
@@ -123,6 +123,34 @@ def add_subscription():
 
 
 @subscription.route(
+    "/redfish/v1/EventService/EventSubscriptions/<id>", methods=["DELETE"])
+def remove_subscription(id):
+    """Removes a specific Subscription
+
+        Args:
+            id: The Subscription ID.
+
+        Returns:
+            Subscription JSON.
+    """
+    try:
+        sc = util.all_subscriptions[id]
+        event_types = sc.redfish["EventTypes"]
+
+        for event in event_types:
+            del util.subscriptions_by_type[event][id]
+
+        del util.all_subscriptions[id]
+
+        return Response(
+            status=status.HTTP_200_OK,
+            mimetype="application/json")
+    except KeyError as e:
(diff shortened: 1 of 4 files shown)


oneview_redfish_toolkit/blueprints/subscription.py
```diff
@@ -122,6 +122,34 @@ def add_subscription():
...
+
+        Returns:
+            Subscription JSON.
+    """
+    try:
+        sc = util.all_subscriptions[id]
+        event_types = sc.redfish["EventTypes"]
+
+        for event in event_types:
+            del util.subscriptions_by_type[event][id]
```
ricardoas: This line also can yield a KeyError, right?

I know it is a corner case... but right now it is impossible to delete a subscription created with:

```
{
  "Destination": "http://www.myserveruri.com",
  "EventTypes": ["Alert", "Alert", "ResourceRemoved"],
  "Context": "1234512344512345"
}
```

Is there anything you can do about it?

oneview_redfish_toolkit/tests/blueprints/test_subscription.py
```diff
@@ -199,6 +199,42 @@ def test_add_subscription_invalid_destination(self, uuid_mockup):
         self.assertEqual(status.HTTP_400_BAD_REQUEST, response.status_code)
         self.assertEqual("application/json", response.mimetype)
 
+    @mock.patch('uuid.uuid1')
+    def test_remove_subscription(self, uuid_mockup):
```
ricardoas: can you check if the subscription still exists after deleting? Maybe if you use on get_subscription...

oneview_redfish_toolkit/blueprints/subscription.py
llvieira: Now I'm treating and removing all duplicated event types in the **POST** operation.

Therefore, if you create a `Subscription` like this:
```javascript
{
  "Destination": "http://www.myserveruri.com",
  "EventTypes": ["Alert", "Alert", "ResourceRemoved"],
  "Context": "1234512344512345"
}
```

The result is going to be exactly like this:

```javascript
{
  "Destination": "http://www.myserveruri.com",
  "EventTypes": ["Alert", "ResourceRemoved"],
  "Context": "1234512344512345"
}
```
So, we no longer have this error when deleting a Subscription.

oneview_redfish_toolkit/tests/blueprints/test_subscription.py
llvieira: Checked! thanks...

llvieira: I also added a new test to check the creation of a subscription with duplicated elements...

PR #241, opened 2018-04-18
Add OV SCMB configuration in the configuration file

### LINKED ISSUE #228: Fix SCMB connection issue
When using authentication_mode “conf” the service does not start anymore.

lhsvobodaj:
Move some OneView SCMB configurations to the Redfish service configuration file.

Closes #228


redfish.conf
```diff
@@ -15,6 +15,10 @@ api_version = 300
 userName = administrator
 password =
 
+[oneview_scmb]
+port = 5671
+socketTimeout = 5
```
danielreed: Why add the SCMB config to the conf file? Most likely these will never need to be changed and by allowing the user to change them it could break the connection to the SCMB.

danielreed: I don't think we want to let the user change it.  If they do we won't be able to connect to the OneView SCMB.  And I don't think HPE will ever change that port number.

Issue #249, opened 2018-05-02
Redfish Events notification


adriane-cardozo:
Send events to Redfish subscribers

Issue #214, opened 2018-01-29
Subscribe to each EventType


ricardoas:
Once communication to HPE OneView SCMB is stablished, the service should subscribe to a topic corresponding to each alert type on OneView (mapping directly to an event type on Redfish.

ffdarkpenguin:
waiting for PR #205 and #208 to be merged

PR #254, opened 2018-05-14
Add support for Redfish EventService API

### LINKED ISSUE #201: How are HPE OneView event types structured?
How are events organized at HPE OneView resource model and how can they be mapped to [Redfish schema](https://redfish.dmtf.org/schemas/Event.v1_2_1.json)?

Check [python-hpOneView](https://github.com/HewlettPackard/python-hpOneView/tree/master/examples/scmb) example code to connect to SCMB and receive events.

### LINKED ISSUE #215: Add TestEvent Action
As oneview-redfish-toolkit user interested in using EventService, I am able to test my event callback server by submitting a POST request to "EventService/Actions/EventService.SubmitTestEvent" triggering a test event to all subscribed users.

lhsvobodaj:
Closes #201 #214 #215 #249

### DIFF
--- a/oneview_redfish_toolkit/api/event.py
+++ b/oneview_redfish_toolkit/api/event.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-# Copyright (2017) Hewlett Packard Enterprise Development LP
+# Copyright (2017-2018) Hewlett Packard Enterprise Development LP
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may
 # not use this file except in compliance with the License. You may obtain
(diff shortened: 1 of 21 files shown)


oneview_redfish_toolkit/api/event.py
```diff
@@ -20,42 +20,79 @@
...
-        event["Severity"] = oneview_alert["resource"]["severity"]
-        event["EventTimestamp"] = oneview_alert["resource"]["created"]
-        self.redfish["Events"].append(event)
+        self.redfish["Events@odata.count"] = 1
+        self.redfish["Events"] = list()
+
+        if (category == "alerts"):
+            self.build_event_from_alert(oneview_message)
+        else:
+            self.build_event_from_task(oneview_message)
```
danielreed: Is the message here actually just a OneView resource and not necessarily a OneView Task?

oneview_redfish_toolkit/api/event.py
```diff
@@ -20,42 +20,79 @@
...
+        event_record = {}
+        event_record["EventTimestamp"] = oneview_alert["timestamp"]
+        event_record["EventType"] = "Alert"
+        event_record["MessageId"] = "Base.1.1.Success"
+        # TODO(svoboda) add link to Redfish resource
+        # event_record["OriginOfCondition"] = origin_of_condition
+
+        self.redfish["Events"].append(event_record)
+
+    def build_event_from_task(self, oneview_task):
```
danielreed: Is the message here really only a task or is it always a resource?

oneview_redfish_toolkit/blueprints/event_service.py
```diff
@@ -16,20 +16,54 @@
...
+        "category": "alerts",
+        "associatedResource": {
+            "resourceName": "0000A66101, bay 3",
+            "resourceUri": "/rest/server-hardware/"
+                           "30373737-3237-4D32-3230-313530314752"
+        }
+    }
+}
+
+ONEVIEW_TEST_TASK = {
```
danielreed: Maybe use resource in the name instead of task.

oneview_redfish_toolkit/util.py
```diff
@@ -408,3 +410,37 @@ def generate_certificate(dir_name, file_name, key_length, key_type="rsa"):
...
+            url = urlparse(subscription.redfish['Destination'])
+            connection = HTTPConnection(url.hostname, port=url.port)
+            json_str = event.serialize()
+
+            connection.request(
+                'POST',
+                url.path,
+                json_str,
+                {'Content-Type': 'application/json'})
+    except ConnectionError as e:
```
adriane-cardozo: We should add retries as specified in the config file.
[event_service]
DeliveryRetryAttempts = 3
DeliveryRetryIntervalSeconds = 30 

Thanks

oneview_redfish_toolkit/api/event.py
lhsvobodaj: At this point, we have an SCMB message, which can be an alert or a task. According to the OneView documentation, the JSON payload of each message from the SCMB contains a `resource` attribute, from which we can get the `category` field value and check whether it is a task or an alert message.

oneview_redfish_toolkit/api/event.py
lhsvobodaj: Here and at the `build_event_from_alert` function, the parameter is always an SCMB message. When it is a task message, we use values from the `resource` object to build the EventRecord. Otherwise, when it is an alert, values from the `associatedResource` are used.

oneview_redfish_toolkit/blueprints/event_service.py
lhsvobodaj: I am not sure whether resource is the best way to describe this variable. What do you think about using `alert` for SCMB messages containing alerts and use the term `message` for "regular" messages? We will end up having something like ONEVIEW_TEST_ALERT and ONEVIEW_TEST_MESSAGE. Any better idea?

oneview_redfish_toolkit/util.py
lhsvobodaj: Done! Please, check the file `event_dispatcher.py`

oneview_redfish_toolkit/api/service_root.py
```diff
@@ -56,9 +56,9 @@ def __init__(self, oneview_uuid):
         self.redfish["CompositionService"] = collections.OrderedDict()
         self.redfish["CompositionService"]["@odata.id"] = \
             "/redfish/v1/CompositionService"
-        # self.redfish["EventService"] = collections.OrderedDict()
-        # self.redfish["EventService"]["@odata.id"] = \
-        # "/redfish/v1/EventService"
+        self.redfish["EventService"] = collections.OrderedDict()
```
adriane-cardozo: Just realized that, if authentication mode is set to "conf", we should not add "EventService" endpoint, since it is not supported now.

lhsvobodaj: You mean, we should only include the EventService API when the authentication mode is set to `conf`, right? I will fix this by adding a test before registering the EventService blueprint in the `app.py`, ok?

adriane-cardozo: Yes, both changes.

PR #289, opened 2018-06-15
Implement Delete for  Composed Resource

### LINKED ISSUE #264: Implement DELETE to delete a composed resource
DELETE request deletes a server profile applied.
Since the toolkit does not support POST yet, a server profile can be created directly in Oneview API e use the UUID to build Redfish URI for testing.

victorhugorodrigues:
Closes #264

### DIFF
--- a/oneview_redfish_toolkit/blueprints/computer_system.py
+++ b/oneview_redfish_toolkit/blueprints/computer_system.py
@@ -185,3 +185,28 @@ def change_power_state(uuid):
         # In case of error log exception and abort
         logging.exception('Unexpected error: {}'.format(e))
         abort(status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+
+@computer_system.route(
+    "/redfish/v1/Systems/<uuid>", methods=["DELETE"])
+def remove_subscription(uuid):
+    """Removes a specific System
+
+        Args:
+            uuid: The System ID.
+    """
+    try:
+        # Deletes server profile for given UUID
+        sucess = g.oneview_client.server_profiles.delete(uuid)
+
+        return Response(
+            status=status.HTTP_200_OK,
+            mimetype="application/json")
+    except HPOneViewException as e:
+        # In case of error log exception and abort
+        logging.exception(e)
+
+        if e.oneview_response['errorCode'] == "RESOURCE_NOT_FOUND":
+            abort(status.HTTP_404_NOT_FOUND, "Computer Sytem not found")
+        else:
+            abort(status.HTTP_500_INTERNAL_SERVER_ERROR)
--- a/oneview_redfish_toolkit/blueprints/computer_system.py
+++ b/oneview_redfish_toolkit/blueprints/computer_system.py
@@ -194,6 +194,13 @@ def remove_subscription(uuid):
 
         Args:
             uuid: The System ID.
+
+        Exceptions:
+            HPOneViewException: When some OneView resource was not found.
+            return abort(404)
+
+            Exception: Unexpected error.
+            return abort(500)
     """
     try:
         # Deletes server profile for given UUID
--- a/oneview_redfish_toolkit/blueprints/computer_system.py
+++ b/oneview_redfish_toolkit/blueprints/computer_system.py
@@ -199,6 +199,9 @@ def remove_subscription(uuid):
         # Deletes server profile for given UUID
         sucess = g.oneview_client.server_profiles.delete(uuid)
 
+        if not sucess:
+            abort(status.HTTP_500_INTERNAL_SERVER_ERROR)
+
         return Response(
             status=status.HTTP_200_OK,
             mimetype="application/json")
--- a/oneview_redfish_toolkit/blueprints/computer_system.py
+++ b/oneview_redfish_toolkit/blueprints/computer_system.py
@@ -200,13 +200,15 @@ def remove_computer_system(uuid):
     response = g.oneview_client.server_profiles.delete(uuid)
 
     if response is True:
-        return Response(status=status.HTTP_204_NO_CONTENT)
+        return Response(status=status.HTTP_204_NO_CONTENT,
+                        mimetype="application/json")
 
     # Check if returned a task
     if type(response) is dict:
         # Check if task is completed
         if response['taskState'] == 'Completed':
-            return Response(status=status.HTTP_204_NO_CONTENT)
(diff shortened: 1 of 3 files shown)


oneview_redfish_toolkit/blueprints/computer_system.py
```diff
@@ -185,3 +185,38 @@ def change_power_state(uuid):
...
+        Exceptions:
+            HPOneViewException: When some OneView resource was not found.
+            return abort(404)
+
+            Exception: Unexpected error.
+            return abort(500)
+    """
+    try:
+        # Deletes server profile for given UUID
+        sucess = g.oneview_client.server_profiles.delete(uuid)
```
lhsvobodaj: success

oneview_redfish_toolkit/blueprints/computer_system.py
```diff
@@ -185,3 +185,38 @@ def change_power_state(uuid):
...
+
+        if not sucess:
```
lhsvobodaj: According to the python SDK, the DELETE request may return a `True` boolean value when the profile has been successfully [removed](https://github.com/HewlettPackard/python-hpOneView/blob/a44fe186f64f4035e6cce939c67e423ecd28f777/hpOneView/resources/resource.py#L269), or a task resource as the response. If we receive a `True` as response, our code will work correctly. However, when the response is a task resource, we need to check its result before sending an OK.

oneview_redfish_toolkit/blueprints/computer_system.py
```diff
@@ -185,3 +185,38 @@ def change_power_state(uuid):
...
+            abort(status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+        return Response(
+            status=status.HTTP_200_OK,
```
lhsvobodaj: In this case, as there is no response body, we could return the status code 204 (No Content)

oneview_redfish_toolkit/tests/blueprints/test_computer_system.py
```diff
@@ -458,3 +458,72 @@ def test_change_power_state_invalid_key(self):
         )
         self.assertEqual("application/json", response.mimetype)
         self.assertEqual(result, invalid_json_key)
+
+    @mock.patch.object(computer_system, 'g')
```
lhsvobodaj: It is a bit odd that this test is working, as we are not setting any behavior in the oneview_client mock. I was expecting to see a `g.oneview_client.server_profiles.delete.return_value = True`. Why is it not necessary?

oneview_redfish_toolkit/blueprints/computer_system.py
```diff
@@ -185,3 +185,38 @@ def change_power_state(uuid):
         # In case of error log exception and abort
         logging.exception('Unexpected error: {}'.format(e))
         abort(status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+
+@computer_system.route(
+    "/redfish/v1/Systems/<uuid>", methods=["DELETE"])
+def remove_subscription(uuid):
```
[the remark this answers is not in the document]
victorhugorodrigues: Done

oneview_redfish_toolkit/blueprints/computer_system.py
victorhugorodrigues: Done

oneview_redfish_toolkit/blueprints/computer_system.py
victorhugorodrigues: Nice input. Now it's handling task as well.

oneview_redfish_toolkit/blueprints/computer_system.py
victorhugorodrigues: Done

oneview_redfish_toolkit/tests/blueprints/test_computer_system.py
victorhugorodrigues: Because I was not checking if the response was True, just handling if it was not False. Once it was a mock object, it was different from False. But now the code is explicitly checking for True.

oneview_redfish_toolkit/blueprints/computer_system.py
```diff
@@ -185,3 +186,50 @@ def change_power_state(uuid):
...
+        if type(response) is dict:
+            # Check if task is completed
+            if response['taskState'] == 'Completed':
+                return Response(status=status.HTTP_204_NO_CONTENT)
+
+            # Log task error message if it has one
+            if response['taskState'] in TASK_ERROR_STATES:
+                if 'taskErrors' in response and \
+                    len(response['taskErrors']) > 0:
+                    err = response['taskErrors'][0]
```
sidneifranco: Can we assume that just a single entry "[0]" will be there? If we could have more, can we assume that the index "0" is our target?

victorhugorodrigues: That is a good point, I've look into python sdk to check how it handle it, and it get just the first entry to get the error message: 
https://github.com/HewlettPackard/python-hpOneView/blob/master/hpOneView/resources/task_monitor.py#L140

But maybe we can concat with other entries.

oneview_redfish_toolkit/blueprints/computer_system.py
```diff
@@ -185,3 +186,33 @@ def change_power_state(uuid):
...
+    """Removes a specific System
+
+        Args:
+            uuid: The System ID.
+    """
+    # Deletes server profile for given UUID
+    response = g.oneview_client.server_profiles.delete(uuid)
+
+    if response is True:
+        return Response(status=status.HTTP_204_NO_CONTENT)
```
ricardogpsf: I dont sure about whether omition of mimetype here works good for all possible clients.
All our application is using `mimetype="application/json"`, but if no problem omiting this, ok. 
The status is "no content", so maybe works good for all situations. =)

oneview_redfish_toolkit/blueprints/computer_system.py
```diff
@@ -185,3 +186,33 @@ def change_power_state(uuid):
...
+
+    # Check if returned a task
+    if type(response) is dict:
+        # Check if task is completed
+        if response['taskState'] == 'Completed':
+            return Response(status=status.HTTP_204_NO_CONTENT)
```
ricardogpsf: Same case here.

PR #292, opened 2018-06-18
Add ResourceBlock for OneView server hardware

### LINKED ISSUE #250: Implement ResourceBlock for Server Hardware
one ResourceBlock for Server Hardware. Map processor and memory information.

Processors[]:
- For each Processor:
  - TotalCores: "ov processorCoreCount"
  - ID: incremental number
  - MaxSpeedMHz: "ov processorSpeedMhz"
  - ProcessorType: CPU

Memory[]:
- ID: 1
- Name: "Memory"
- CapacityMiB: "ov memoryMb"

Response from OV:
  "powerState": "On",
  "processorCoreCount": 4,
  "processorCount": 2,
  "processorSpeedMhz": 2400,
  "processorType": "Intel(R) Xeon(R) CPU E5620 @ 2.40GHz", 
  "memoryMb": 1048576

lhsvobodaj:
* Create specific API class for resource block of type server hardware
 * Include OneView and Redfish mockups for testing purposes
 * Create API class for a Computer System of a Resource Block

Closes #250


oneview_redfish_toolkit/tests/blueprints/test_resource_block.py
```diff
@@ -0,0 +1,218 @@
...
+        response = self.app.get(
+            "/redfish/v1/CompositionService/ResourceBlocks"
+            "/30303437-3034-4D32-3230-313133364752/Systems/1234567890")
+
+        self.assertEqual(status.HTTP_404_NOT_FOUND, response.status_code)
+        self.assertEqual("application/json", response.mimetype)
+
+    @mock.patch.object(resource_block, 'g')
+    def test_get_computer_system(self, g):
+        with open(
```
ricardogpsf: This block is used by more than one test, right?! We could put it inside the setUp or out of the test class (but in the same file) to reuse it.

oneview_redfish_toolkit/blueprints/resource_block.py
```diff
@@ -0,0 +1,152 @@
...
+
+from hpOneView.exceptions import HPOneViewException
+from oneview_redfish_toolkit.api.errors import OneViewRedfishError
+from oneview_redfish_toolkit.api.resource_block_computer_system \
+    import ResourceBlockComputerSystem
+from oneview_redfish_toolkit.api.server_hardware_resource_block \
+    import ServerHardwareResourceBlock
+
+
+RESOURCE_BLOCK_URI = "/redfish/v1/CompositionService/ResourceBlocks"
```
ricardogpsf: This constant is very good, but is referenced just by this class and this URI `"/redfish/v1/CompositionService/ResourceBlocks"` is being used by api class, too. Maybe it should be inside the api class and referenced here (or keep inside here, but referenced in api class, I prefer the first option =).

oneview_redfish_toolkit/blueprints/resource_block.py
```diff
@@ -0,0 +1,152 @@
...
+
+        # Build response and returns
+        response = Response(
+            response=json_str,
+            status=status.HTTP_200_OK,
+            mimetype="application/json")
+        response.headers.add("ETag", "W/" + eTag)
+
+        return response
+    except HPOneViewException as e:
```
victorhugorodrigues: Could be interesting we check different error codes from HPOneViewException to return different responses(404, 500, ...), or can we assume the exception is a 404?

oneview_redfish_toolkit/blueprints/resource_block.py
```diff
@@ -0,0 +1,152 @@
...
+
+        json_str = computer_system.serialize()
+
+        response = Response(
+            response=json_str,
+            status=status.HTTP_200_OK,
+            mimetype="application/json")
+        response.headers.add("ETag", "W/" + server_hardware["eTag"])
+        return response
+    except HPOneViewException as e:
```
victorhugorodrigues: Same question above

oneview_redfish_toolkit/tests/blueprints/test_resource_block.py
lhsvobodaj: done

oneview_redfish_toolkit/blueprints/resource_block.py
lhsvobodaj: done

oneview_redfish_toolkit/tests/blueprints/test_resource_block.py
victorhugorodrigues: Since it's not used in all test cases, place it on setUp will make it runs this block for every test case. Maybe the better way it's to place it on setUpClass, but in order to use setUpClass on children classes from TestBase, we need to refactor TestBase class. For this refactoring maybe we can use setUpModule to place the code that today is on the setUp from TestBase.

oneview_redfish_toolkit/blueprints/resource_block.py
lhsvobodaj: I agree. This fine-grained error handling will be much easier after we adopt the ResponseBuilder, as all the logic to handle different SDK exceptions can be defined in only one place.

oneview_redfish_toolkit/tests/blueprints/test_resource_block.py
```diff
@@ -0,0 +1,211 @@
...
+from oneview_redfish_toolkit.blueprints import resource_block
+from oneview_redfish_toolkit.tests.base_test import BaseTest
+
+
+class TestResourceBlock(BaseTest):
+    """Tests for ResourceBlock blueprint"""
+
+    @classmethod
+    def setUpClass(self):
+        # creates a test client
```
lhsvobodaj: super.setUpClass

Issue #295, opened 2018-06-18
Update Systems collection


adriane-cardozo:
System collection should list only server profiles applied.
Remove Physical systems (implemented in the phase 1).

PR #315, opened 2018-06-28
Update Systems collection

karolcatunda:
- Removing physical systems from system collection
- Updating computer system unit tests to get new system collection data
Closes: https://github.com/HewlettPackard/oneview-redfish-toolkit/issues/295


oneview_redfish_toolkit/api/computer_system_collection.py
```diff
@@ -23,28 +23,30 @@ class ComputerSystemCollection(RedfishJsonValidator):
     """Creates a Computer System Collection Redfish dict
 
         Populates self.redfish with some hardcoded ComputerSystemCollection
-        values and with the response of Oneview with all server hardware.
+        values and with the response of Oneview with all servers profiled
+        applied.
```
sidneifranco: maybe it could be better to change a little bit to:"... with all servers with Server Profile applied"

oneview_redfish_toolkit/blueprints/computer_system_collection.py
```diff
@@ -48,8 +48,15 @@ def get_computer_system_collection():
             raise OneViewRedfishResourceNotFoundError(
                 "server-hardware-list", "Resource")
 
+        # Filter server profiles that has a profile applied
```
sidneifranco: Filter servers that have a profile applied

oneview_redfish_toolkit/mockups/oneview/ServerHardwaresProfiledApplied.json
```diff
@@ -0,0 +1,1165 @@
+[{"processorSpeedMhz":2400,
```
sidneifranco: Consider renaming the file to ServersHardwareProfiled.json or something like that.

oneview_redfish_toolkit/api/computer_system_collection.py
```diff
@@ -59,10 +61,10 @@ def _set_redfish_members(self):
             Populates self.redfish["Members"] with the links to Redfish
             ComputerSystems.
         """
-        for server_hardware, index in \
-                zip(self.server_hardware,
-                    range(len(self.server_hardware))):
-
+        for server_profile, index in \
```
ricardogpsf: just a note: Could we do this iteration using a simple loop `for` without apply a `zip` method to the `self.server_profiles` and without `index` variable? This will keep the code more simple, no? What do you think?

For example:
```python
for item in self.server_profiles:
    dict_result = {"@odata.id": item["serverProfileUri"].split("/")[-1]}
    self.redfish["Members"].append(dict_result)
```

oneview_redfish_toolkit/blueprints/computer_system_collection.py
```diff
@@ -48,8 +48,15 @@ def get_computer_system_collection():
             raise OneViewRedfishResourceNotFoundError(
```
ricardogpsf: I know this behavior is previously than that current PR, but we should not raise exception when the list is empty, just return the empty list with status 200. 
We can do that right now. What do you think? @karolcatunda @victorhugorodrigues @adriane-cardozo @lhsvobodaj

oneview_redfish_toolkit/blueprints/computer_system_collection.py
```diff
@@ -48,8 +48,15 @@ def get_computer_system_collection():
...
                 "server-hardware-list", "Resource")
 
+        # Filter server profiles that has a profile applied
+        server_profiles_applied_list = list()
+        for server_hardware_item, index in \
+                zip(server_hardware_list, range(len(server_hardware_list))):
```
ricardogpsf: Could we use simples loop `for` as explained before in another comment?

oneview_redfish_toolkit/api/computer_system_collection.py
```diff
@@ -23,28 +23,30 @@ class ComputerSystemCollection(RedfishJsonValidator):
...
         Populates self.redfish with some hardcoded ComputerSystemCollection
-        values and with the response of Oneview with all server hardware.
+        values and with the response of Oneview with all servers profiled
+        applied.
     """
 
     SCHEMA_NAME = 'ComputerSystemCollection'
 
-    def __init__(self, server_hardware):
+    def __init__(self, server_profiles):
```
ricardogpsf: Seeing the blueprint code I believe that server_profiles name is bad in this case because this list is a server_hardware list, in fact, right? So, we can named as server_hardware.

oneview_redfish_toolkit/blueprints/computer_system_collection.py
victorhugorodrigues: I agree.

oneview_redfish_toolkit/blueprints/computer_system_collection.py
victorhugorodrigues: We could go ahead and change the name of the list as well.

oneview_redfish_toolkit/blueprints/computer_system_collection.py
adriane-cardozo: Yes, we should not raise exception for empty collections. Anyway, I think we should retrieve the list of server profiles directly instead of checking for each server hardware. Is there a reason why not doing that?

oneview_redfish_toolkit/mockups/oneview/ServerHardwaresProfiledApplied.json
adriane-cardozo: Can we use ServerProfiles.json or ServerProfilesApplied.json? To follow Oneview naming structure. Let's discuss tomorrow.

oneview_redfish_toolkit/blueprints/computer_system_collection.py
adriane-cardozo: Need to discuss how to handle profiles that are not assigned to a server hardware... maybe that's the reason :) Let's discuss tomorrow.

karolcatunda:
All comments were fixed on last commit. Please, check them @sidneifranco @ricardogpsf @adriane-cardozo

PR #319, opened 2018-06-29
Include Storage resource blocks in a Resource Zone

### LINKED ISSUE #299: Include Storage resource blocks in a Resource Zone

### LINKED ISSUE #309: Storage Resource Blocks in a Zone should be available only if storage controller is configured
For a Zone, Storage resource blocks should be listed in the Resource Blocks collection if the template has storage controller configured.

ricardogpsf:
- Storage Resource Blocks are available only if storage controller is configured and is not Embedded
- Adds unit tests to cover the scenario above

Ps.: This PR closes 2 issues (#309  and #299), but that issues will be put on CHANGELOG only after the Sprint to avoid merge problems.

Closes #299 
Closes #309

### DIFF
--- a/oneview_redfish_toolkit/api/zone.py
+++ b/oneview_redfish_toolkit/api/zone.py
@@ -32,7 +32,7 @@ class Zone(RedfishJsonValidator):
 
     SCHEMA_NAME = 'Zone'
 
-    def __init__(self, profile_template, available_targets_obj):
+    def __init__(self, profile_template, available_targets_obj, drives):
         """Zone constructor
 
             Populates self.redfish with the contents of
@@ -45,7 +45,15 @@ class Zone(RedfishJsonValidator):
         """
         super().__init__(self.SCHEMA_NAME)
 
-        available_targets = available_targets_obj["targets"]
+        self.available_targets = available_targets_obj["targets"]
+        self.drives = []
+
+        controllers = profile_template["localStorage"]["controllers"]
+        has_not_embedded = \
+            [i for i in controllers if i["deviceSlot"] != "Embedded"]
+
+        if has_not_embedded:
+            self.drives = drives
 
         self.redfish["@odata.type"] = "#Zone.v1_1_0.Zone"
         self.redfish["Id"] = profile_template["uri"].split("/")[-1]
@@ -56,7 +64,7 @@ class Zone(RedfishJsonValidator):
         self.redfish["Links"] = dict()
         self.redfish["Links"]["ResourceBlocks"] = list()
 
-        self.fill_resource_blocks(profile_template, available_targets)
+        self.fill_resource_blocks()
 
         self.capabilities_key = "@Redfish.CollectionCapabilities"
         self.redfish[self.capabilities_key] = dict()
@@ -72,9 +80,12 @@ class Zone(RedfishJsonValidator):
 
         self._validate()
 
-    def fill_resource_blocks(self, profile_template, available_targets):
-        for item in available_targets:
+    def fill_resource_blocks(self):
+        for item in self.available_targets:
             self.add_resource_block_item_to_links(item, "serverHardwareUri")
+
+        for item in self.drives:
+            self.add_resource_block_item_to_links(item, "uri")
 
     def add_resource_block_item_to_links(self, original_dict, uri_key):
         uuid = original_dict[uri_key].split("/")[-1]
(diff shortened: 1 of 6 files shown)


oneview_redfish_toolkit/api/zone.py
```diff
@@ -32,7 +32,7 @@ class Zone(RedfishJsonValidator):
 
     SCHEMA_NAME = 'Zone'
 
-    def __init__(self, profile_template, available_targets_obj):
+    def __init__(self, profile_template, available_targets_obj, drives):
```
karolcatunda: I believe is missing driver doc on Args.

oneview_redfish_toolkit/tests/api/test_zone.py
```diff
@@ -31,29 +31,50 @@ class TestZone(BaseTest):
...
-            zone = Zone(self.server_profile_template, self.available_targets)
-        except Exception as e:
-            self.fail("Failed to instantiate Zone class."
-                      " Error: {}".format(e))
-
-        try:
-            result = json.loads(zone.serialize())
-        except Exception as e:
-            self.fail("Failed to serialize. Error: {}".format(e))
+        zone = Zone(self.server_profile_template,
```
karolcatunda: Test doc is missing.

oneview_redfish_toolkit/tests/api/test_zone.py
```diff
@@ -31,29 +31,50 @@ class TestZone(BaseTest):
...
+        profile_template["localStorage"]["controllers"] = []
+
+        zone = Zone(profile_template,
+                    self.available_targets,
+                    self.drives)
+        result = json.loads(zone.serialize())
+
+        self.assertEqual(self.zone_without_drives_mockup, result)
+
+    def test_drives_as_links_when_storage_controller_is_embedded(self):
```
karolcatunda: Here too.

oneview_redfish_toolkit/tests/api/test_zone.py
```diff
@@ -31,29 +31,50 @@ class TestZone(BaseTest):
...
-        except Exception as e:
-            self.fail("Failed to serialize. Error: {}".format(e))
+        zone = Zone(self.server_profile_template,
+                    self.available_targets,
+                    self.drives)
+        result = json.loads(zone.serialize())
 
         self.assertEqual(self.zone_mockup, result)
+
+    def test_drives_as_links_when_storage_controllers_are_not_configured(
```
karolcatunda: Same comment above :+1:

oneview_redfish_toolkit/api/zone.py
```diff
@@ -45,7 +45,15 @@ def __init__(self, profile_template, available_targets_obj):
         """
         super().__init__(self.SCHEMA_NAME)
 
-        available_targets = available_targets_obj["targets"]
+        self.available_targets = available_targets_obj["targets"]
```
victorhugorodrigues: Is there a reason to this be an instance attribute and not just a variable? I see it is used inside private methods, but it could be passed as argument, not creating code dependency. Now, methods like "fill_resource_blocks()" depends that instance variable "self.available_targets" has be set before call the method.

oneview_redfish_toolkit/api/zone.py
```diff
@@ -72,10 +80,13 @@ def __init__(self, profile_template, available_targets_obj):
 
         self._validate()
 
-    def fill_resource_blocks(self, profile_template, available_targets):
-        for item in available_targets:
+    def fill_resource_blocks(self):
+        for item in self.available_targets:
```
victorhugorodrigues: Could this method receives "available_target" as argument? Not depending the instance variable has been set before.

oneview_redfish_toolkit/tests/api/test_zone.py
ricardogpsf: Do you know is necessary for this case of tests?

oneview_redfish_toolkit/api/zone.py
ricardogpsf: No special reason, I has used instance attributes because the method that are using them should be a non public method, so should be called only inside the class and is being called by the constructor that know how prepare the instance variable correctly.
But, I can do the requested changes. =)

oneview_redfish_toolkit/api/zone.py
ricardogpsf: Yes, I dont see problem with instance variables in this case, but I will do that.

PR #322, opened 2018-06-29
Migrate to stestr unit test library

### LINKED ISSUE #321: Review build dependencies
We should set a version for each lib to not update automatically.

To fix build, Ricardo also test other changes, so we should check if we need them: https://github.com/ricardoas/oneview-redfish-toolkit/tree/fix-unit-test

ricardoas:
@adriane-cardozo as I said, I took a look on the latest changes on unit test facilities in http://github.com/openstack/ironic and noticed they migrated to stestr unit test runner. I performed the same migration the the build turned ok again. As you already fix it in another way (updating oslo), I will leave this here only for reference in case you need to perform this migration in a near future.

PS: on the upper-constraints reference on tox.ini, I am really sorry for that! When starting the project, we used https://github.com/openstack-dev/cookiecutter as a template base as it was the best option at the moment. So we stripped (almost) everything openstack related... my bad!

Closes #321

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ cover/
 nosetests.xml
 .testrepository
 .venv
+.stestr
 
 # Translations
 *.mo
--- a/.stestr.conf
+++ b/.stestr.conf
@@ -1,0 +1,3 @@
+[DEFAULT]
+test_path=${TESTS_DIR:-./oneview_redfish_toolkit/tests/}
+top_dir=./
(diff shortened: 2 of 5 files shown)


adriane-cardozo:
Closes #321

ricardogpsf:
The problem was solved before by @lhsvobodaj and, for now, we decide don't migrate to `stestr` lib. After, we can back to this PR if is the case.
Thanks for this PR and your suggestion. 😃

PR #327, opened 2018-07-03
Include Network resource blocks in a Resource Zone

### LINKED ISSUE #294: Include Network resource blocks in a Resource Zone
Update resource blocks:  remove template entry and add Network resource blocks (one for each template).

victorhugorodrigues:
Closes #294

### DIFF
--- a/oneview_redfish_toolkit/api/zone.py
+++ b/oneview_redfish_toolkit/api/zone.py
@@ -91,7 +91,7 @@ class Zone(RedfishJsonValidator):
     def _fill_network_resource_block(self, profile_template):
         conn_settings = profile_template["connectionSettings"]
 
-        if len(conn_settings["connections"]):
+        if conn_settings["connections"]:
             dict_item = dict()
             dict_item["@odata.id"] = ResourceBlockCollection.BASE_URI + \
                 "/" + profile_template["uri"].split("/")[-1]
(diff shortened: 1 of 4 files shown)


oneview_redfish_toolkit/api/zone.py
```diff
@@ -87,6 +88,15 @@ def fill_resource_blocks(self, available_targets, drives):
         for item in drives:
             self.add_resource_block_item_to_links(item, "uri")
 
+    def _fill_network_resource_block(self, profile_template):
+        conn_settings = profile_template["connectionSettings"]
+
+        if len(conn_settings["connections"]):
```
ricardogpsf: As the `conn_settings["connections"]` is a list, you can check the conditional statement without calling the `len` method, you have the same result.
```python
 if conn_settings["connections"]:
```
But it is not so important, if you want do that, good. If not, no problem. =)

oneview_redfish_toolkit/api/zone.py
```diff
@@ -65,6 +65,7 @@ def __init__(self, profile_template, available_targets_obj, drives=[]):
         self.redfish["Links"]["ResourceBlocks"] = list()
 
         self.fill_resource_blocks(available_targets, drives)
+        self._fill_network_resource_block(profile_template)
```
adriane-cardozo: It seems this call should be executed inside fill_resource_blocks() that already handles hardware and drives.

PR #345, opened 2018-07-12
Network Resource Block: add support for network sets

### LINKED ISSUE #339: Network Resource Block: add support for network sets
Need to map EthernetInterface when connection in the template is related to a network set instead of a network. For a network set, VLANs array should be returned.

victorhugorodrigues:
Closes #339


oneview_redfish_toolkit/blueprints/vlan_network_interface.py
```diff
@@ -0,0 +1,114 @@
...
+
+    try:
+        ethernet_network = \
+            g.oneview_client.ethernet_networks.get(vlan_id)
+
+        vlan = VLanNetworkInterface(ethernet_network, request.path)
+
+        return ResponseBuilder.success(vlan)
+
+    except OneViewRedfishError as e:
```
karolcatunda: Can OneViewRedfishError type exception occur here? I'm asking because it doesn't seem be raised in this method, because if any exception happens, flask will catch through response builder.

victorhugorodrigues: Done.

PR #354, opened 2018-07-20
Refactoring redfish.conf

### LINKED ISSUE #349: Refactor Redfish.conf
- Remove schema_dir and registry_dir configuration that should not be changed by the user. Update code to retrieve from global var .
- Remove sections [schemas] and [registry].
- Remove  api_version configuration and update README (it works with OneView 4.0 and above).
- Remove duplicate Redfish.conf.
- Update installer
- Investigate integration with CherryPy.

victorhugorodrigues:
Closes #349


oneview_redfish_toolkit/blueprints/metadata.py
```diff
@@ -39,8 +39,8 @@ def get_metadata():
     """
 
     try:
-        schemas = collections.OrderedDict(util.config.items('schemas'))
-        mtdt = Metadata(schemas)
+        schemas_dict = collections.OrderedDict(schemas.SCHEMAS)
```
ricardogpsf: why do we using a OrderedDict here? Could we use a normal dict?

victorhugorodrigues: I did'nt want to change the logic of application on this refatoring so I keep it as OrderedDict. But I'm not sure if we need this.

ricardogpsf: 😄 👍

PR #367, opened 2018-07-30
Authentication for each request should be based on the REDFISH_SESSION_TOKEN defined during Redfish session creation

### LINKED ISSUE #267: Authentication for each request should be based on the REDFISH_SESSION_TOKEN defined during Redfish session creation
Supporting multiple OneView instances, during Redfish session creation, the service should create as many OneView sessions as the number of OneView instances and assign one of them (can be the first OneView session token) as the Redfish session token that will be returned to the Redfish client.

Changes to blueprints/session.py and app.py/check_authentication().

victorhugorodrigues:
Closes #267

### DIFF
--- a/oneview_redfish_toolkit/authentication.py
+++ b/oneview_redfish_toolkit/authentication.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-# Copyright (2017-2018) Hewlett Packard Enterprise Development LP
+# Copyright (2018) Hewlett Packard Enterprise Development LP
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may
 # not use this file except in compliance with the License. You may obtain
--- a/oneview_redfish_toolkit/authentication.py
+++ b/oneview_redfish_toolkit/authentication.py
@@ -36,6 +36,14 @@ def get_map_tokens():
     return globals()['map_tokens']
 
 
+def set_map_tokens(new_map):
+    globals()['map_tokens'] = new_map
+
+
+def set_new_token(redfish_token, tokens_ov_by_ip):
+    globals()['map_tokens'][redfish_token] = tokens_ov_by_ip
+
+
 def login(username, password):
     try:
         tokens_ov_by_ip = dict()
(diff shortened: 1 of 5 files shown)


oneview_redfish_toolkit/authentication.py
```diff
@@ -0,0 +1,95 @@
...
+
+# Modules own libs
+from oneview_redfish_toolkit import config
+from oneview_redfish_toolkit import connection
+
+
+# Globals vars:
+#   globals()['map_tokens']
+
+def get_map_tokens():
```
ricardogpsf: Could I assert that the public methods of this `authentication.py` are `login` and ` check_authentication`?
If yes, we could keep the others methods as private.

oneview_redfish_toolkit/authentication.py
```diff
@@ -0,0 +1,95 @@
...
+    return globals()['map_tokens']
+
+
+def set_map_tokens(new_map):
+    globals()['map_tokens'] = new_map
```
karolcatunda: Isn't better to assign a empty dict here rather than receive it as argument?

oneview_redfish_toolkit/authentication.py
victorhugorodrigues: Done.

oneview_redfish_toolkit/tests/test_authentication.py
```diff
@@ -0,0 +1,63 @@
...
+
+        # Check if redfish token return is one of the OneView's token
+        rf_token = authentication.login('user', 'password')
+        self.assertTrue(rf_token in list_tokens)
+
+        # Check if cached token map has the Redfish token return on login
+        map_tokens = authentication.get_map_tokens()
+        self.assertTrue(rf_token in map_tokens)
+
+        # Check if chaced token map has the correct OneViewIp/OneViewToken
```
karolcatunda: cached* :smile:

oneview_redfish_toolkit/authentication.py
victorhugorodrigues: Done. Maybe after multiple OneView support implementation we could move this initialization method to run at application startup.

PR #376, opened 2018-07-31
Updates GET capabilities object for handle the new ID

### LINKED ISSUE #368: Update GET capabilities object for the new ID
URI for a capabilities object can be "templateUUID>-<enclosureID>"

ricardogpsf:
- The new ID could be only the SPT Id ou SPT Id + Enclosure Id

Closes #368

### DIFF
--- a/oneview_redfish_toolkit/blueprints/computer_system.py
+++ b/oneview_redfish_toolkit/blueprints/computer_system.py
@@ -39,12 +39,13 @@ from oneview_redfish_toolkit.api.redfish_json_validator \
 from oneview_redfish_toolkit.api.util.power_option import OneViewPowerOption
 from oneview_redfish_toolkit.blueprints.util.response_builder import \
     ResponseBuilder
+from oneview_redfish_toolkit.blueprints import zone
 
 computer_system = Blueprint("computer_system", __name__)
 
 
-@computer_system.route("/redfish/v1/Systems/<uuid>", methods=["GET"])
-def get_computer_system(uuid):
+@computer_system.route("/redfish/v1/Systems/<system_id>", methods=["GET"])
+def get_computer_system(system_id):
     """Get the Redfish Computer System for a given UUID.
 
         Return ComputerSystem redfish JSON for a given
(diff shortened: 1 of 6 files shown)


oneview_redfish_toolkit/blueprints/computer_system.py
```diff
@@ -58,12 +59,13 @@ def get_computer_system(uuid):
     """
 
     try:
-        resource = _get_oneview_resource(uuid)
+        resource_id, encl_id = zone.split_base_id_and_enclosure_id(uuid)
```
adriane-cardozo: I think we should rename "uuid" by "id" at  @computer_system.route("/redfish/v1/Systems/<id>", methods=["GET"])

oneview_redfish_toolkit/blueprints/computer_system.py
```diff
@@ -58,12 +59,13 @@ def get_computer_system(uuid):
...
+        resource = _get_oneview_resource(resource_id)
         category = resource["category"]
 
         if category == 'server-profile-templates':
-            computer_system_resource = CapabilitiesObject(resource)
-        elif category == 'server-profiles':
+            computer_system_resource = CapabilitiesObject(uuid, resource)
```
adriane-cardozo: This will impact POST to computer system since the Id will be <template_uuid>-<enclosureid>. Please update blueprints/computer_system.py line 225: spt = g.oneview_client.server_profile_templates.get(body["Id"]) to retrieve the template.
Also api/computer_system.py method build_server_profile should save the id in the Description and not template uri, also where Description is read should be changed.

adriane-cardozo: As agreed, this will be fixed in another PR.

oneview_redfish_toolkit/blueprints/computer_system.py
ricardogpsf: I will change to `system_id` to avoid shadowing of built-in name "`id`" of python.

PR #385, opened 2018-08-07
Fixing Storage for a Composed System when there are no drives

### LINKED ISSUE #382: Fix Storage for Composed System when there are no drives
It is returning 500, but the profile does not need to have any drives.

https://{{redfish-ip}}/redfish/v1/Systems/145d5f6b-837b-428c-829d-ea40dcfe3f03/Storage/1

karolcatunda:
Closes #382

### DIFF
--- a/oneview_redfish_toolkit/api/storage.py
+++ b/oneview_redfish_toolkit/api/storage.py
@@ -94,7 +94,8 @@ class Storage(RedfishJsonValidator):
 
         count_drives_by_jbod = \
             [int(item["numPhysicalDrives"]) for item in sas_logical_jbods]
-        count_drives = reduce(operator.add, count_drives_by_jbod)
+
+        count_drives = reduce(operator.add, count_drives_by_jbod, 0)
 
         self.redfish["Drives@odata.count"] = count_drives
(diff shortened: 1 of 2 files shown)


oneview_redfish_toolkit/api/storage.py
```diff
@@ -92,9 +92,12 @@ def __init__(self,
...
 
-        count_drives_by_jbod = \
-            [int(item["numPhysicalDrives"]) for item in sas_logical_jbods]
-        count_drives = reduce(operator.add, count_drives_by_jbod)
+        count_drives = 0
+        if len(sas_logical_jbods) > 0:
+            count_drives_by_jbod = \
+                [int(item["numPhysicalDrives"]) for item in sas_logical_jbods]
+
+            count_drives = reduce(operator.add, count_drives_by_jbod)
```
ricardogpsf: Just a tip, it is not a necessary change, we could simply start the initial value of `reduce` method with zero: `reduce(operator.add, count_drives_by_jbod, 0)`, so the `count_drives` will be `0` when the `sas_logical_jbods` is empty.

PR #394, opened 2018-08-13
Implementing Service Root endpoint without authenticate

### LINKED ISSUE #393: Credentials should not be required in production mode
Credentials in the conf should be used only demo purpose and not required for execution.

victorhugorodrigues:
Closes #393

### DIFF
--- a/oneview_redfish_toolkit/connection.py
+++ b/oneview_redfish_toolkit/connection.py
@@ -108,22 +108,7 @@ def check_oneview_availability(oneview_ip):
 
     for attempt_counter in range(attempts):
         try:
-            connection = HTTPSConnection(
-                oneview_ip, context=ssl.SSLContext(ssl.PROTOCOL_TLSv1_2))
-
-            connection.request(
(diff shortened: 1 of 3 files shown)


oneview_redfish_toolkit/connection.py
```diff
@@ -138,14 +123,38 @@ def check_oneview_availability(oneview_ip):
...
                 time.sleep(retry_interval_sec)
-        finally:
-            connection.close()
 
     message = "After {} attempts OneView is unreachable at {}".format(
         attempts, oneview_ip)
     raise OneViewRedfishError(message)
 
 
+def request_oneview(oneview_ip, rest_url):
```
ricardogpsf: `test_url` means that is a url to test a connection with oneview? Why not name it just as url?

victorhugorodrigues: It's the rest url to be requested on OneView, it will be used to test OneView connection and to get appliance node information for service root endpoint as well.

PR #411, opened 2018-08-20
Changes POST to compose a system to return the new uuid as soon as profile uuid is available

### LINKED ISSUE #375: Change POST to compose a system to return the new uuid as soon as profile uuid is available
POST should return the new uuid and not wait until the profile creation is complete (current implementation). Use task for monitoring until server profile uuid is available.
oneview_client.connection.post ("url", server_profile)
Loop for task monitoring (5 seconds) until task returns the server profile URI.

ricardogpsf:
- Uses task for monitoring server profile until the uuid is available.

Closes #375

### DIFF
--- a/oneview_redfish_toolkit/handler_multiple_oneview.py
+++ b/oneview_redfish_toolkit/handler_multiple_oneview.py
@@ -59,7 +59,6 @@ RESOURCE_STRATEGY = {
         },
     "server_hardware_types": {"get": st.first_parameter_resource},
     "server_profiles": {
-        "create": st.create_server_profile,
         "delete": st.delete_server_profile,
         "get": st.first_parameter_resource,
         "get_available_targets": st.first_parameter_resource,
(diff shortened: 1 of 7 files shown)


oneview_redfish_toolkit/handler_multiple_oneview.py
```diff
@@ -24,7 +24,10 @@
 
 RESOURCE_STRATEGY = {
     "appliance_node_information": {"get_version": st.all_oneviews_resource},
-    "connection": {"get": st.first_parameter_resource},
+    "connection": {
+        "get": st.first_parameter_resource,
+        "post": st.create_server_profile,
+        },
```
victorhugorodrigues: Maybe it's better to remove the mapped strategy "create_server_profile" for "server_profiles.create" from this same dict.
Once it's being changed how the strategy(create_server_profile) handles the arguments, so even that we are not using right now, if we call "g.oneview_client.server_profiles.create()" it will have an error.

PR #412, opened 2018-08-20
Handling Concurrency for cached data

### LINKED ISSUE #366: Concurrency for multiple OneView
Handling concurrency for multiple OneView support on:
- Map for Redfish -> OneView authentication tokens;
- Map for OneView resource -> OneView client.

victorhugorodrigues:
Closes #366

### DIFF
--- a/oneview_redfish_toolkit/authentication.py
+++ b/oneview_redfish_toolkit/authentication.py
@@ -17,6 +17,7 @@
 # Python libs
 import logging
 import logging.config
+from threading import Lock
 
 # 3rd party libs
 from flask import abort
@@ -33,6 +34,7 @@ from oneview_redfish_toolkit import connection
 # Globals vars:
 #   globals()['map_tokens']
 
+
 def _get_map_tokens():
     return globals()['map_tokens']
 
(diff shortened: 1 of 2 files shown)


oneview_redfish_toolkit/authentication.py
```diff
@@ -42,7 +44,12 @@ def init_map_tokens():
 
 
 def _set_new_token(redfish_token, tokens_ov_by_ip):
-    globals()['map_tokens'][redfish_token] = tokens_ov_by_ip
+    lock = Lock()
```
ricardogpsf: Why do not use `with lock` instead?
The code would be:
```python
with lock:
    globals()['map_tokens'][redfish_token] = tokens_ov_by_ip
```
This statement, implicitly, will apply `lock.acquire()` and `lock.release()` as is said [here](https://docs.python.org/3/library/threading.html#using-locks-conditions-and-semaphores-in-the-with-statement).

victorhugorodrigues: You are right, I want to implement it in a more explicit way, but we sure can change it.

ricardogpsf: what do you think?
I want you to feel at ease. I prefer another way because of simplicity.
But, the code has not problem for me. We already can merge it if all agree, to. =)

PR #413, opened 2018-08-20
Zone name must be unique

### LINKED ISSUE #391: Zone name must be unique
We must concatenate "enclosure X" in the zone name when there are more than one zone for a SPT

karolcatunda:
- Adding a implementation to verify if Zone has a enclosure. If it has enclosure id will be added into name property on the Zone.
 - Fixing mockups, adding enclosure name into Name property

Closes #391

### DIFF
--- a/oneview_redfish_toolkit/api/zone.py
+++ b/oneview_redfish_toolkit/api/zone.py
@@ -53,7 +53,7 @@ class Zone(RedfishJsonValidator):
                 server_hardware_list: Oneview's server hardware list
                 (servers and empty bays) for assignment to a server profile
                 enclosure_name: Enclosure name associated with a Zone.
-                If Zone has not enclosure it will be empty.
+                If Zone has not enclosure it will be None.
                 drives: Oneview's dict drives list
         """
         super().__init__(self.SCHEMA_NAME)
--- a/oneview_redfish_toolkit/api/zone.py
+++ b/oneview_redfish_toolkit/api/zone.py
@@ -63,7 +63,8 @@ class Zone(RedfishJsonValidator):
 
         self.redfish["@odata.type"] = self.get_odata_type()
         self.redfish["Id"] = resource_id
-        self.redfish["Name"] = profile_template["name"]
+        self.redfish["Name"] = self.get_zone_name(
+            profile_template["name"], resource_id)
         status_from_ov = profile_template["status"]
         self.redfish["Status"] = status_mapping.STATUS_MAP[status_from_ov]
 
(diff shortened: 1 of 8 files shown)


oneview_redfish_toolkit/api/zone.py
```diff
@@ -87,6 +88,15 @@ def __init__(self,
 
         self._validate()
 
+    def get_zone_name(self, spt_name, resource_id):
+        zone_name = spt_name
+        resource_id_group = resource_id.split("-")
```
ricardogpsf: Here we are not concatenating the enclosure name, but the id, maybe this information will not be friendly user as a Zone name, we could use the enclosure name.

oneview_redfish_toolkit/api/zone.py
```diff
@@ -87,6 +88,15 @@ def __init__(self,
...
+        # if resource group len is more than 5
+        # it means that exists an enclosure
+        if len(resource_id_group) > 5:
```
ricardogpsf: In both cases (using enclosure name or enclosure id) we can avoid doing this code by receiving the enclosure name (or enclosure id) as a parameter of constructor, because we already have that [information](https://github.com/HewlettPackard/oneview-redfish-toolkit/blob/master/oneview_redfish_toolkit/blueprints/zone.py#L41) in zone blueprint.

oneview_redfish_toolkit/tests/blueprints/test_zone.py
```diff
@@ -125,6 +125,10 @@ def test_get_zone_when_uuid_is_only_template_id(self):
         ) as f:
             zone_without_drives_mockup = json.load(f)
 
+        zone_name_only_template_id = \
```
ricardogpsf: The name of this variable seems weird because the name has not template_id, maybe should be `zone_name_only_with_template_name `, but all right with this, don't need change it.

oneview_redfish_toolkit/api/zone.py
```diff
@@ -51,6 +52,8 @@ def __init__(self,
                 profile_template: Oneview's server profile template dict
                 server_hardware_list: Oneview's server hardware list
                 (servers and empty bays) for assignment to a server profile
+                enclosure_name: Enclosure name associated with a Zone.
+                If Zone has not enclosure it will be empty.
```
ricardogpsf: "It will be empty"? Or "None"?

oneview_redfish_toolkit/tests/api/test_zone.py
```diff
@@ -77,6 +79,7 @@ def test_drives_as_links_when_storage_controllers_are_not_configured(self):
         zone = Zone(zone_id,
                     profile_template,
                     self.server_hardware_list,
+                    self.enclosure_name_none,
```
ricardogpsf: Why dont pass simply `None`? Instead of declare a new variable to assign `None`.

PR #420, opened 2018-08-24
ZoneCollection is listing Zones even for enclosures that do not have Drive Enclosure

### LINKED ISSUE #392: ZoneCollection is listing Zones even for enclosures that do not have Drive Enclosure
Should fix _get_enclosures_uris_by_template to check if the enclosure has a drive enclosure, if not it should not be listed as a Zone.

karolcatunda:
- Adding a feature to check if an enclosure has drive enclosures associated with. If it has not, it will not be listed on the zone collection
 - Adding a new strategy on multiple oneview file to get all drive enclosures that uses or nor a filter and fixing strategies that receives a filter, supporting string or list filter types
 - Updating unit tests affected by these changes and creating a new one to test it

Closes #392


oneview_redfish_toolkit/strategy_multiple_oneview.py
```diff
@@ -101,3 +109,22 @@ def _run_action(resource_id, resource_get, function_get, resource,
     return multiple_oneview.query_ov_client_by_resource(resource_id,
                                                         resource, function,
                                                         *args, **kwargs)
+
+
+def _get_resource_id_by_filter(kwargs, resource_uri):
```
victorhugorodrigues: Maybe "filter_parameter", "filter_property" or "property_name" would fit better for this argument.

oneview_redfish_toolkit/services/zone_service.py
```diff
@@ -97,16 +99,40 @@ def get_zone_ids_by_templates(self, server_profile_templates):
...
             else:
                 zone_ids.append(template_id)
 
         return zone_ids
+
+    def _get_valid_logical_enclosures(self):
+        all_logical_enclosures_uris = self._get_all_enclosures_uris()
+        valid_logical_enclosures = list()
+        for logical_enclosure_uri in all_logical_enclosures_uris:
+            logical_encl_has_drives = self.ov_client.drive_enclosures.get_all(
```
ricardogpsf: In another [point of project](https://github.com/HewlettPackard/oneview-redfish-toolkit/blob/master/oneview_redfish_toolkit/blueprints/zone.py#L75-L78) we are getting information about drive enclosures associated to the enclosure, too.
We could verify what is the best way to this current case, because the result here comes with all drives too, resulting to a possible very large data returned and we just want verify if drive enclosure is present.
Could you verify the best approach, please? Let me know about it. 😃

karolcatunda: The approach suggested by you returns more data that I will use. The request that I do only returns a 'members' list that is the drive enclosures associated or not with the enclosure filtered. If there's no drive enclosure, the list will be empty. So it is more simple to check the data in this case than in yours. But I appreciate your suggestion 👍

oneview_redfish_toolkit/services/zone_service.py
```diff
@@ -97,16 +97,35 @@ def get_zone_ids_by_templates(self, server_profile_templates):
                 server_profile_templates: the list of Server Profile Template
         """
         zone_ids = []
+        expected_enclosures_uris = self._get_enclosures_uris()
```
adriane-cardozo: To make it more clear, please rename to "enclosure_uris_with_drives" or "enclosure_uris_with_valid_drive_enclosures". The method also could be renamed to "_get_enclosures_uris_with_valid_drive_enclosures" for example.

PR #424, opened 2018-08-28
Change Oneview Redfish Toolkit to use CherryPy

### LINKED ISSUE #348: Change Oneview Redfish Toolkit to use CherryPy

ricardogpsf:
- Changes the app.py code to use CherryPy standard;
- Adds TransLogger to handler logs between cherrypy and flask app
- Uses Daemonizer plugin of the cherrypy to run the app as a daemon process
- Changes the requirements.txt adding the cherrypy to be installed

Closes #348

### DIFF
--- a/oneview_redfish_toolkit/app.py
+++ b/oneview_redfish_toolkit/app.py
@@ -24,7 +24,6 @@ from threading import Thread
 # 3rd party libs
 import cherrypy
 from cherrypy.process.plugins import Daemonizer
-from cherrypy.process.plugins import PIDFile
 from flask import abort
 from flask import Flask
 from flask import g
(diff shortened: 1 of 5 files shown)


oneview_redfish_toolkit/app.py
```diff
@@ -369,8 +373,38 @@ def hp_oneview_client_exception(exception):
...
+        'server.socket_port': port,
+        'server.socket_host': host,
+        'server.ssl_certificate': ssl_cert_file,
+        'server.ssl_private_key': ssl_key_file
+    })
+
+    app_logged = TransLogger(app.wsgi_app, setup_console_handler=False)
+    cherrypy.tree.graft(app_logged, '/')
+
+    Daemonizer(cherrypy.engine).subscribe()
```
ricardogpsf: I removed the code that creates pid file because is not so important, I had added it before to make it easier to recover the process PID, but with the new code we would gain more complexity too.

ricardogpsf:
I am working now changing the code to use cherrypy for both production and development environments.

ricardogpsf:
> I am working now changing the code to use cherrypy for both production and development environments.
>
> This will be good to avoid run the app in a server different that the production server and to have an unexpected behavior in production environment due that.
>
>Thanks for this suggestion @sidneifranco .

Done!

PR #426, opened 2018-08-30
Use a label to save template_UUID instead of Description attribute in the server profile

### LINKED ISSUE #370: Use a label to save template_UUID instead of Description attribute in the server profile

karolcatunda:
- Creating a new handle to post a new label associated with a server profile using server profile template description as label
 - Replacing server profile template description from server profile to use labels associated with a server profile
 - Fixing unit tests to use labels instead server profile description

Closes #370

### DIFF
--- a/oneview_redfish_toolkit/api/computer_system.py
+++ b/oneview_redfish_toolkit/api/computer_system.py
@@ -45,7 +45,7 @@ class ComputerSystem(RedfishJsonValidator):
     BASE_URI = '/redfish/v1/Systems'
 
     def __init__(self, server_hardware, server_hardware_types,
-                 server_profile, drives):
+                 server_profile, drives, server_profile_labels):
         """ComputerSystem constructor
 
             Populates self.redfish with the contents of ServerHardware and
@@ -56,6 +56,7 @@ class ComputerSystem(RedfishJsonValidator):
                 server_hardware_types: ServerHardwareTypes dict from OneView
                 server_profile: ServerProfile dict from OneView.
                 drives: Drives list from OneView
+                server_profile_labels: Server Profile labels from Oneview
         """
         super().__init__(self.SCHEMA_NAME)
 
@@ -113,9 +114,9 @@ class ComputerSystem(RedfishJsonValidator):
         self.redfish["Links"]["ManagedBy"][0]["@odata.id"] = \
             "/redfish/v1/Managers/" + server_hardware['uuid']
         self.redfish["Links"]["ResourceBlocks"] = list()
-        self._fill_resource_block_members(base_resource,
-                                          drives,
-                                          server_hardware)
+        self._fill_resource_block_members(drives,
+                                          server_hardware,
+                                          server_profile_labels)
         self.redfish["Actions"] = collections.OrderedDict()
         self.redfish["Actions"]["#ComputerSystem.Reset"] = \
             collections.OrderedDict()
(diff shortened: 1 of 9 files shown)


oneview_redfish_toolkit/api/computer_system.py
```diff
@@ -236,30 +237,29 @@ def _build_sas_logical_jbods(server_profile_template, storage_blocks):
...
+                                  server_hardware,
+                                  server_profile_labels):
         resource_block_uuids = list()
         resource_block_uuids.append(server_hardware["uuid"])
 
         # fill with network resource block
-        if server_profile["description"]:
-            spt_id = server_profile["description"].split("/")[-1]
-            resource_block_uuids.append(spt_id)
+        spt_id = server_profile_labels["labels"][0]["name"].replace(" ", "-")
```
victorhugorodrigues: Once we can have ServerProfiles created directly on OneView without using Redfish Toolkit, I think we need to check if it has just the one expected label on the expected format, to make sure that this label was created by the toolkit and represent and ServerProfileTemplate.

oneview_redfish_toolkit/mockups/oneview/ServerProfileWithLabels.json
```diff
@@ -0,0 +1,15 @@
+{
```
ricardogpsf: The name of this file ("ServerProfileWithLabels") is not representing the correct resource, because the resource is not a ServerProfile, it is a `"type":"ResourceLabels"`. Maybe we could name it as "LabelForServerProfile".
What do you think?

oneview_redfish_toolkit/api/computer_system.py
ricardogpsf: And we should put a test covering this case, once is a valid flux.

adriane-cardozo: Since the profile is created via toolkit, I think the first label will always be the one we created even if someone uses OneView to create other labels to the server profile later. Please can you just check this behavior? Anyway, I agree that we should confirm if there is a label (someone/script can remove it) and if it represents a serverprofiletemplate (someone/script can change it - I think) , and if not do not add the link.

karolcatunda: Yes @adriane-cardozo and @victorhugorodrigues  . You're correct. If we created via toolkit, the first label will always be create by using the spt on the post request, independent if there are other labels created on Oneview. But I agree that we need to check if the value passed represents in fact a spt :+1: .

oneview_redfish_toolkit/mockups/oneview/ServerProfileWithLabels.json
karolcatunda: Seems a great suggestion. I will rename it.

ricardogpsf: Rename the variable names representing it mockup, too. =)

oneview_redfish_toolkit/blueprints/computer_system.py
```diff
@@ -324,3 +329,25 @@ def _get_drives_from_sp(server_profile):
...
+    for label in all_sp_labels["labels"]:
+        try:
+            spt_uuid = label["name"].replace(" ", "-")
+            is_valid_spt = \
+                g.oneview_client.server_profile_templates.get(spt_uuid)
+            if is_valid_spt:
+                server_profile_template_uuid = spt_uuid
+                break
+        except HPOneViewException as e:
+            if e.oneview_response["errorCode"] == 'RESOURCE_NOT_FOUND':
```
victorhugorodrigues: We could use NOT_FOUND_ONEVIEW_ERRORS array from "api/errors.py" instead.

oneview_redfish_toolkit/blueprints/computer_system.py
```diff
@@ -324,3 +330,25 @@ def _get_drives_from_sp(server_profile):
             jbods_drives.extend(drives_by_sas_logical_uuid)
 
     return jbods_drives
+
+
+def _get_server_profile_template_from_sp(sp_uri):
```
ricardogpsf: What do you think about move this method to the [ComputerSystemService](https://github.com/HewlettPackard/oneview-redfish-toolkit/blob/master/oneview_redfish_toolkit/services/computer_system_service.py#L24)?

PR #427, opened 2018-08-30
Adds support for authLoginDomain

### LINKED ISSUE #421: Add support for authLoginDomain in the Redfish.conf

ricardogpsf:
- Supports auth authLoginDomain using both conf and session mode.

Closes #421

### DIFF
--- a/oneview_redfish_toolkit/authentication.py
+++ b/oneview_redfish_toolkit/authentication.py
@@ -106,8 +106,7 @@ def create_credentials(username, password):
     try:
         login_domain, username = username.split("\\")
     except (AttributeError, ValueError):
-        logging.debug("Trying build credentials to login to Oneview, but "
-                      "userName '{}' has not authLoginDomain".format(username))
+        logging.debug("AuthLoginDomain not specified on session creation")
     else:
         credentials["authLoginDomain"] = login_domain
 
--- a/oneview_redfish_toolkit/authentication.py
+++ b/oneview_redfish_toolkit/authentication.py
@@ -53,7 +53,6 @@ def login(username, password):
     try:
         tokens_ov_by_ip = dict()
         credentials = create_credentials(username, password)
-
         for ip in config.get_oneview_multiple_ips():
             ov_config = \
                 connection.create_oneview_config(ip=ip,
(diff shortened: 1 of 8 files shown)


oneview_redfish_toolkit/authentication.py
```diff
@@ -103,6 +102,15 @@ def get_multiple_oneview_token():
 
 def create_credentials(username, password):
     credentials = dict()
+
+    try:
+        login_domain, username = username.split("\\")
+    except (AttributeError, ValueError):
+        logging.info("Trying build credentials to login to Oneview, but "
+                     "userName '{}' has not authLoginDomain".format(username))
```
victorhugorodrigues: Maybe we could use debug message instead of info message.

oneview_redfish_toolkit/conf/redfish.conf
```diff
@@ -12,6 +12,7 @@ ip =
 [credentials]
 userName =
 password =
+authLoginDomain =
```
adriane-cardozo: Shouldn't we set "LOCAL" as default?

oneview_redfish_toolkit/authentication.py
```diff
@@ -103,6 +102,15 @@ def get_multiple_oneview_token():
 
 def create_credentials(username, password):
     credentials = dict()
+
+    try:
+        login_domain, username = username.split("\\")
+    except (AttributeError, ValueError):
```
adriane-cardozo: I would change the message to something like  "AuthLoginDomain not specified on session creation", otherwise it seems an error.

oneview_redfish_toolkit/conf/redfish.conf
adriane-cardozo: Or does OneView handle no value on this attribute?

ricardogpsf: We can set "LOCAL" as default yes! I had set a empty value just to keep the same default value used for all keys in the `[credentials]` session in that file `redfish.conf` (all values are empty in that session).

adriane-cardozo: Ok

PR #429, opened 2018-09-03
Add support for Description in Composed system and Capabilities Object

### LINKED ISSUE #387: Add support for Description in Composed system and Capabilities Object
Add Description attribute in the Computer System resource.
Depends on #370

victorhugorodrigues:
Closes #387

### DIFF
--- a/oneview_redfish_toolkit/api/computer_system.py
+++ b/oneview_redfish_toolkit/api/computer_system.py
@@ -63,6 +63,8 @@ class ComputerSystem(RedfishJsonValidator):
         base_resource = server_profile
         self.redfish["@odata.type"] = self.get_odata_type()
         self.redfish["Id"] = base_resource["uuid"]
+        if base_resource["description"]:
+            self.redfish["Description"] = base_resource["description"]
         self.redfish["Name"] = base_resource["name"]
         self.redfish["SystemType"] = "Composed"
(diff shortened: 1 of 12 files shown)


oneview_redfish_toolkit/api/computer_system.py
```diff
@@ -63,6 +63,8 @@ def __init__(self, server_hardware, server_hardware_types,
         base_resource = server_profile
         self.redfish["@odata.type"] = self.get_odata_type()
         self.redfish["Id"] = base_resource["uuid"]
+        if base_resource["description"]:
```
adriane-cardozo: Description is a string, since it is supported it should be listed even when it is empty.

victorhugorodrigues: Done.

oneview_redfish_toolkit/blueprints/computer_system.py
```diff
@@ -222,6 +222,7 @@ def create_composed_system():
 
         server_profile = ComputerSystem.build_server_profile(
             body["Name"],
+            body.get("Description"),
```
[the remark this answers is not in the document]
victorhugorodrigues: the get function is able to handle if the "Description" key does not exists, it will return None, for the cases that the user doesn't define any Description for the composed system.

PR #439, opened 2018-09-10
Fix unit tests for conf mode and client session tests

victorhugorodrigues:
-Fix unit tests for conf mode
-Fix client session tests


ricardogpsf:
Looks good. All test are passing, too (using both authentication_mode "conf" and "session").
:shipit:

PR #442, opened 2018-09-12
Feature/garbage collector for expired sessions

### LINKED ISSUE #441: Remove abandoned and expired sessions
It can occurs a case when the client is not using the session and abandoned it while it is working well, but our application only remove when the authorization error is raised by a request to Oneview. 

So, we need crate a routine to remove old sessions not used and expired.

ricardogpsf:
- Starts a Thread that every 12 hours looks for expired and abandoned sessions tokens, and remove them from cache.

Closes #441 

### How test the solution? ###
You can change the expiration time of a specific session in Oneview, just apply a request like below (changing the ONEVIEW_IP and ONEVIEW_TOKEN to valid values):
```bash
curl -X POST \
  https://ONEVIEW_IP/rest/sessions/idle-timeout \
  -H 'auth: ONEVIEW_TOKEN' \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/json' \
  -H 'session-id: ONEVIEW_TOKEN' \
  -H 'x-api-version: 600' \
  -d '{ "idleTimeout": 60000 }'
```
The `idleTimeout` is the value in **milliseconds** to keep alive the session.
For our example above, 60000 is the time to keep alive during 1 minute.

Steps to test:
1) Create more than one session on Redfish server.
2) List the Redfish sessions accessing `/redfish/v1/SessionService/Sessions`
3) Post a request to change the session `idleTimeout` to the Oneview server
4) Wait the time to expire Oneview session based on idleTimeout
5) List the Redfish sessions again. You should see one session less than previously.

### DIFF
--- a/oneview_redfish_toolkit/api/errors.py
+++ b/oneview_redfish_toolkit/api/errors.py
@@ -20,7 +20,8 @@ NOT_FOUND_ONEVIEW_ERRORS = ['RESOURCE_NOT_FOUND', 'ProfileNotFoundException',
 
 AUTH_ONEVIEW_ERRORS = ['AUTHN_AUTH_FAIL',
                        'AUTHN_AUTH_FAIL_LOGINDOMAINNOTFOUND',
-                       'AUTHORIZATION']
+                       'AUTHORIZATION',
+                       'Session.INVALID']
 
 
 class OneViewRedfishError(Exception):
--- a/oneview_redfish_toolkit/client_session.py
+++ b/oneview_redfish_toolkit/client_session.py
@@ -63,13 +63,16 @@ def _gc_for_expired_sessions():
     map_items = _get_map_clients().items()
     for token, dict_by_token in map_items:
         ov_clients = iter(dict_by_token['client_ov_by_ip'].values())
-        ov_client = next(ov_clients)
+        ov_client = next(ov_clients)  # ov_clients can be in any order
+        ov_session_id = ov_client.connection.get_session_id()
         try:
             # request that does not increment the life of Oneview session
-            resp, body = ov_client.connection.do_http('GET',
-                                                      '/rest/sessions/',
-                                                      '',
-                                                      {'Session-Id': token})
+            resp, body = ov_client.connection.do_http(
+                'GET',
+                '/rest/sessions/',
+                '',
+                {'Session-Id': ov_session_id}
+            )
 
             # when is not a success
             if resp.status not in range(200, 300):
--- a/oneview_redfish_toolkit/client_session.py
+++ b/oneview_redfish_toolkit/client_session.py
@@ -18,7 +18,9 @@
 from collections import OrderedDict
 import logging
 import logging.config
+import threading
 from threading import Lock
+import time
 import uuid
 
 # 3rd party libs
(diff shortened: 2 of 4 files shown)


oneview_redfish_toolkit/api/errors.py
```diff
@@ -20,7 +20,8 @@
 
 AUTH_ONEVIEW_ERRORS = ['AUTHN_AUTH_FAIL',
                        'AUTHN_AUTH_FAIL_LOGINDOMAINNOTFOUND',
-                       'AUTHORIZATION']
+                       'AUTHORIZATION',
+                       'Session.INVALID']
```
victorhugorodrigues: We don't have an OneView documentation for all 401 OneView errors right?

oneview_redfish_toolkit/client_session.py
```diff
@@ -44,6 +49,49 @@ def init_map_clients():
     globals()['map_clients'] = OrderedDict()
 
 
+def init_gc_for_expired_sessions():
+    gc_thread = threading.Thread(target=_gc_for_expired_sessions)
+    gc_thread.daemon = True
```
victorhugorodrigues: You could set this as a instantiation parameter: t = Thread(daemon=True)

oneview_redfish_toolkit/client_session.py
```diff
@@ -44,6 +49,49 @@ def init_map_clients():
...
+                                  'of Oneview sessions endpoint: {}'
+                                  .format(resp.status, body))
+
+        except Exception as e:
+            logging.exception('Unexpected error: {}'.format(e))
+
+    for token in tokens_to_remove:
+        clear_session_by_token(token)
+
+    _gc_for_expired_sessions()
```
victorhugorodrigues: I think that it's better use and while loop, because if the same method is called repeatedly in an recursive way without never finish the previous execution, the program execution will have to allocate memory for the local variables for each recursive call. Then since the old method execution doesn't finish and instead of that it's calling a new method execution, the allocated memory for the local variable from the previous method execution can not be released, so it can cause a memory leakage.

oneview_redfish_toolkit/client_session.py
```diff
@@ -44,6 +49,49 @@ def init_map_clients():
...
+def _gc_for_expired_sessions():
+    time.sleep(GC_FREQUENCY_IN_SEC)
+    logging.info('Verifying expired sessions...')
+
+    tokens_to_remove = []
+    map_items = _get_map_clients().items()
+    for token, dict_by_token in map_items:
+        ov_clients = iter(dict_by_token['client_ov_by_ip'].values())
+        ov_client = next(ov_clients)  # ov_clients can be in any order
+        ov_session_id = ov_client.connection.get_session_id()
```
victorhugorodrigues: It's checking for just one OneView for each token session right? I'm thinking on multiple OneView scenario where one of them have different settings for expiration time. Do you think that we need to cover this scenario @adriane-cardozo ?

oneview_redfish_toolkit/client_session.py
ricardogpsf: yeaah! OMG, I forgot the problem with the stack of call. :sweat:
Thanks. =)

oneview_redfish_toolkit/client_session.py
ricardogpsf: I imaginate that oneviews should be configured with same expiration time.
I think checking just one client just be sufficient, because all should expire in similar time.

oneview_redfish_toolkit/api/errors.py
ricardogpsf: I dont know. =)
I found these errors by manual tests.

oneview_redfish_toolkit/client_session.py
adriane-cardozo: Yes, let's assume all OV have the same expiration time for now.

PR #443, opened 2018-09-13
Update Managers collection and Manager

### LINKED ISSUE #297: Update Managers collection and Manager
We should have only one Manager - Oneview (Type: service). Need to check other mappings.
We should map all status from controller.json to Redfish Manager enum values.
Collection will not use python sdk to list all managers.
Remove all other Managers implemented in phase 1.
Update Links to ManagedBy for Chassis and ComputerSystem.

Check if we should keep firmware version (appliance node info - software version as used for enclosure).

We should get appliance uuid for the manager uuid GET /rest/appliance/nodeinfo/version).

Depends on #268 (Closed)

karolcatunda:
- Removing server hardware and enclosure queries from ManagerCollection. Now it will query info about the appliances defined on conf file
- Removing all Enclosuremanager and BladeManager references (blueprints, apis and unit tests)
- Refactoring manager to get info about a specific appliance
- Fixing unit tests and created new ones to this new approach

Closes #297


oneview_redfish_toolkit/api/manager.py
```diff
@@ -27,31 +32,51 @@ class Manager(RedfishJsonValidator):
...
+            "/redfish/v1/Managers/" + oneview_appliance_info['uuid']
+
+    @staticmethod
+    def _get_highest_health_state(health_state_members):
+        health_states = dict()
+
+        for member in health_state_members:
+            redfish_health_state = status_mapping.MANAGER_HEALTH_STATE_MAPPING.\
+                get(member["severity"])
+            health_states[redfish_health_state] = \
```
adriane-cardozo: Would this work and simplify if we replace by:

highest_status = \
            max(highest_status, status_mapping.CRITICALITY_STATUS_MAPPING[redfish_health_state])

Seems we can rename all mapping structures that are already in status_mapping, e.g. "CRITICALITY_STATUS"

oneview_redfish_toolkit/blueprints/manager.py
```diff
@@ -91,3 +83,31 @@ def get_managers(uuid):
...
+
+    if not appliance:
+        raise OneViewRedfishError("Could not find manager with id " + uuid)
+
+    return appliance, appliance_index
+
+
+def get_current_manager():
+    oneview_appliances = \
+        g.oneview_client.appliance_node_information.get_version()
```
adriane-cardozo: We must avoid this overhead for each ResourceBlock. I believe we should cache appliance_uuid with the OV_ip during login and this will simplify a lot.

oneview_redfish_toolkit/api/manager.py
karolcatunda: @adriane-cardozo I understand your suggestion for highest_status but the health_state_members list (with 3 members) is composed by the values that we need first, map to the redfish state, as also check between these three values which of them has the most critical state. Because of that, we map each member (severity) to redfish state and then I assigned them to health_states list so that in this way I can check what is the highest status.

Regarding his second remark, I will rename all mapping structures removing "MAPPING" word for each map that has it.

oneview_redfish_toolkit/api/manager.py
```diff
@@ -27,31 +32,54 @@ class Manager(RedfishJsonValidator):
...
+        self.redfish["Status"]["State"] = state
+        self.redfish["Status"]["Health"] = health
         self.redfish["@odata.context"] = \
             "/redfish/v1/$metadata#Manager.Manager"
         self.redfish["@odata.id"] = \
-            "/redfish/v1/Managers/" + oneview_resource['uuid']
+            "/redfish/v1/Managers/" + oneview_appliance_info['uuid']
+
+    @staticmethod
+    def _get_highest_health_state(health_state_members):
```
ricardogpsf: I have a suggestion, let me know if it works well for this situation and what is your opinion:
```python
def _get_highest_health_state(health_state_members):
    health_state_result = None
    highest_status = 0
    
    for member in health_state_members:
        redfish_health_state = status_mapping.MANAGER_HEALTH_STATE.\
                get(member["severity"])
        current_status = status_mapping.CRITICALITY_STATUS[redfish_health_state]

        if current_status > highest_status:
            highest_status = current_status
            health_state_result = redfish_health_state
    
    return health_state_result
```

It is because we can leverage the "for loop" to get the result desired.

oneview_redfish_toolkit/blueprints/manager.py
```diff
@@ -91,3 +82,26 @@ def get_managers(uuid):
...
+            appliance_index = index
+            break
+
+    if not appliance:
+        raise OneViewRedfishError("Could not find manager with id " + uuid)
+
+    return appliance, appliance_index
+
+
+def get_current_manager():
```
ricardogpsf: My suggestion is put this helper methods inside a service file (in services folder) to avoid referencing this blueprint in another blueprint files.

For example this method `get_current_manager` is called in others blueprint files.

With this we are organizing the helper codes and services in a specific files and folder to do that.

oneview_redfish_toolkit/tests/api/test_manager.py
```diff
@@ -0,0 +1,68 @@
+# -*- coding: utf-8 -*-
+
+# Copyright (2017-2018) Hewlett Packard Enterprise Development LP
```
ricardogpsf: (~2017~-2018)
It is a new file, right?! =)

oneview_redfish_toolkit/tests/api/test_manager.py
```diff
@@ -0,0 +1,68 @@
...
+
+        # Loading ApplianceHealthStatusList mockup result
+        with open(
+                'oneview_redfish_toolkit/mockups/oneview/'
+                'ApplianceHealthStatusList.json'
+        ) as f:
+            self.appliance_health_status_list = json.load(f)
+
+            # Loading Manager mockup result
+            with open(
```
ricardogpsf: This indentation is weird.

oneview_redfish_toolkit/blueprints/manager.py
ricardogpsf: Leveraging this thought, we can use `memoization` for this case, for example using [this](https://docs.python.org/3/library/functools.html#functools.lru_cache).

oneview_redfish_toolkit/api/manager.py
karolcatunda: I rollback this code to use lambda again, because the "for" present on lambda is more elegant than the way you suggest. It avoids to create more code and simplify this into one line :+1: 

But I appreciate your suggestion :)

Let's see what @adriane-cardozo and @victorhugorodrigues think about this.

victorhugorodrigues: The way @ricardogpsf suggested is a more explicit way and for me easier to read. But the way you did is a more elegant way using some fancy code. In terms of performance, the way you did the computer execution will have to iterate twice to reach the result, meanwhile the way Ricardo proposed will just iterate once, so I would choose the the way Ricardo proposed.

oneview_redfish_toolkit/services/manager_service.py
```diff
@@ -0,0 +1,25 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+
+# Own libs
+from oneview_redfish_toolkit import multiple_oneview
+
+
+def get_current_manager():
+    map_resources = multiple_oneview.get_map_resources()
+    manager_uuid = list(map_resources.keys())[0]
```
adriane-cardozo: ?

oneview_redfish_toolkit/blueprints/manager.py
```diff
@@ -41,41 +40,32 @@ def get_managers(uuid):
         /redfish/v1/Managers/id is requested.
 
         Returns:
-            JSON: JSON with Managers info for Enclosure or ServerHardware.
+            JSON: JSON with Managers info.
     """
     try:
-        appliance_information = \
+        oneview_appliances_version = \
```
adriane-cardozo: Since we have the uuid in the cache, why do we still need to retrieve from all OneViews?

oneview_redfish_toolkit/blueprints/manager.py
```diff
@@ -41,41 +40,32 @@ def get_managers(uuid):
...
-            enclosure = g.oneview_client.enclosures.get(uuid)
-            etag = enclosure['eTag']
-            manager = EnclosureManager(enclosure, oneview_version)
-        else:
-            raise OneViewRedfishError('Enclosure type not found')
+        ov_appliance_info, appliance_index = \
+            _get_appliance_by_uuid(uuid, oneview_appliances_version)
+
+        state_url = "/controller-state.json"
+        ov_appliances_state = g.oneview_client.connection.get(state_url)
```
adriane-cardozo: same here, why do we need to retrieve from all OneViews?

oneview_redfish_toolkit/blueprints/manager.py
```diff
@@ -41,41 +40,32 @@ def get_managers(uuid):
...
+
+        ov_health_state_url = "/rest/appliance/health-status"
+        ov_appliances_health_status = \
```
adriane-cardozo: same here, why do we need to retrieve from all OneViews?

oneview_redfish_toolkit/services/manager_service.py
victorhugorodrigues: We discussed and @karolcatunda will change to get the OneView IP for the appliance uuid, and make a direct query on this OneView.

oneview_redfish_toolkit/services/manager_service.py
```diff
@@ -0,0 +1,37 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+
+# Own libs
+from oneview_redfish_toolkit import multiple_oneview
+
+
+def get_manager_uuid(resource_id):
+    map_resources = multiple_oneview.get_map_resources()
+    ov_ip = map_resources[resource_id]
```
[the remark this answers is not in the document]
karolcatunda: @adriane-cardozo we don't need to do this check because this method is only called after the resource_id filled in cache, as you can see on this [example](https://github.com/HewlettPackard/oneview-redfish-toolkit/blob/d34afd3ded5a3ef49443ab340426b4a159191038/oneview_redfish_toolkit/blueprints/chassis.py#L57). Here the uuid request is already mapped, so I can search for it on cache, as happens on other calls like this.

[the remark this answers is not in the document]
adriane-cardozo: Got it!

oneview_redfish_toolkit/blueprints/manager.py
```diff
@@ -41,41 +43,36 @@ def get_managers(uuid):
...
+        ov_client = client_session.get_oneview_client(ov_ip)
+
+        ov_appliance_info = multiple_oneview.execute_query_ov_client(
+            ov_client, "appliance_node_information", "get_version"
+        )
+        ov_appliance_state = multiple_oneview.execute_query_ov_client(
+            ov_client, "connection", "get", state_url
+        )
+        ov_appliance_health_status = multiple_oneview.execute_query_ov_client(
+            ov_client, "connection", "get", ov_health_status_url
```
victorhugorodrigues: Nice direct calls reusing the methods on multiple_oneview, this way we keep centralized all OneView request.
Maybe in the future we could keep in request scope which OneView client that request is using for our GET request in a single resource, not for collection. This change could help performance on our GET request for just one resource.

PR #451, opened 2018-09-18
Support for Delete Session (logout)

### LINKED ISSUE #409: Support for Delete Session

ricardogpsf:
Closes #409

### DIFF
--- a/oneview_redfish_toolkit/client_session.py
+++ b/oneview_redfish_toolkit/client_session.py
@@ -99,9 +99,12 @@ def _set_new_client_by_token(redfish_token, client_ov_by_ip):
         }
 
 
-def _get_session_id_by_token(token):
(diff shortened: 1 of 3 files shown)


oneview_redfish_toolkit/client_session.py
```diff
@@ -115,6 +115,14 @@ def get_session_ids():
     return [sess_dict['session_id'] for _, sess_dict in session_map_items]
 
 
+def get_session_by_token(token):
```
adriane-cardozo: Why not use existing method  def _get_session_id_by_token(token)? Maybe rename existing to get_session_id_by_token(token).

ricardogpsf:
Very good pointed @adriane-cardozo . :sweat_smile:

PR #463, opened 2018-09-26
Change order of scmb certs generation

### LINKED ISSUE #401: Change order of scmb certs generation
- try to get scmb certs from OneView and if not available, then generate (just need to change the current order - line 91 before 81).
- also  on app.py when it fails to connect, try to download/generate again

karolcatunda:
- Getting Oneview cert and if it is not found, generating
- On scmb certificate path validation, we are checking also if it is valid

Closes #401


oneview_redfish_toolkit/api/scmb.py
```diff
@@ -61,14 +62,18 @@ def _scmb_key_path():
...
-def check_cert_exist():
+def has_valid_certificates():
     try:
         _scmb_base_dir()
     except KeyError as error:
         logging.error("Invalid configuration for ssl cert. "
                       "Verify the [ssl] section in config file")
         raise error
 
+    return _has_scmb_certificate() and is_cert_working_with_scmb()
```
victorhugorodrigues: Since **_has_scmb_certificate()** internally will call **_scmb_base_dir()**, maybe we could put just this line inside the "**try**" block.

oneview_redfish_toolkit/api/scmb.py
```diff
@@ -101,7 +106,35 @@ def get_cert():
 
     with open(_oneview_ca_path(), 'w+') as f:
         f.write(cert)
-    # Generate scmb Cert:
+
+    global certs
```
victorhugorodrigues: Do we need to declare a global variable?

oneview_redfish_toolkit/api/scmb.py
```diff
@@ -101,7 +106,35 @@ def get_cert():
...
+            logging.info('Generating new SCMB certs in Oneview...')
+            _generate_scmb_certificate(ov_client)
+            certs = ov_client.certificate_rabbitmq.get_key_pair(
+                'default')
+        else:
+            logging.exception("Unexpected error")
+            raise
+
+    logging.info('SCMB certs already generated in Oneview. '
+                 'Getting certs...')
```
victorhugorodrigues: This log message should be in the end of the try block

oneview_redfish_toolkit/app.py
```diff
@@ -315,10 +315,11 @@ def hp_oneview_client_exception(exception):
...
     if config.auth_mode_is_conf():
         # Loading scmb connection
-        if scmb.check_cert_exist():
-            logging.info('SCMB certs already exists testing connection...')
+        if scmb.has_valid_certificates():
+            logging.info('SCMB certs already exists and is valid...')
         else:
-            logging.info('SCMB certs not found. Generating/getting certs....')
+            logging.info('SCMB certs not found. '
+                         'Checking if it was already generated in Oneview...')
```
victorhugorodrigues: We could place all this **if config.auth_mode_is_conf():** code block as a new function **init_event_service** inside scmb.py file. This way we could create new unit tests for this code block,  since we do not test the app.py

oneview_redfish_toolkit/app.py
```diff
@@ -315,10 +315,11 @@ def hp_oneview_client_exception(exception):
...
             scmb.get_cert()
             logging.info('Got certs. Testing connection...')
         if not scmb.is_cert_working_with_scmb():
```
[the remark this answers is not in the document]
karolcatunda: Yes @ricardogpsf . I agree with your suggestion. I will moving this line to the else statement. Thanks :+1:

oneview_redfish_toolkit/api/scmb.py
```diff
@@ -61,14 +66,37 @@ def _scmb_key_path():
...
+                exit(1)
+        scmb_thread = threading.Thread(target=listen_scmb)
+        scmb_thread.daemon = True
+        scmb_thread.start()
+    else:
+        logging.warning("Authentication mode set to session. SCMB events will "
+                        "be disabled")
+
+
+def has_valid_certificates():
```
ricardogpsf: This method is a internal method? We could put it as private method: `_has_valid_certificates()`

oneview_redfish_toolkit/api/scmb.py
```diff
@@ -61,14 +66,37 @@ def _scmb_key_path():
...
     try:
-        _scmb_base_dir()
+        return _has_scmb_certificates_path() and is_cert_working_with_scmb()
```
ricardogpsf: `is_cert_working_with_scmb` is internal method, too? A private method is good for this situation.

oneview_redfish_toolkit/api/scmb.py
```diff
@@ -101,7 +129,33 @@ def get_cert():
...
+
+    # Save cert
+    with open(_scmb_cert_path(), 'w+') as f:
+        f.write(certs['base64SSLCertData'])
+    # Save key
+    with open(_scmb_key_path(), 'w+') as f:
+        f.write(certs['base64SSLKeyData'])
+
+
+def get_scmb_certificate_from_ov(ov_client):
```
ricardogpsf: Is a private method, right? So should be `_get_scmb_certificate_from_ov`.

oneview_redfish_toolkit/api/scmb.py
```diff
@@ -101,7 +129,33 @@ def get_cert():
...
+        if e.oneview_response["errorCode"] in NOT_FOUND_ONEVIEW_ERRORS:
+            logging.info('Generating SCMB cert in Oneview...')
+            generate_certificate_in_oneview(ov_client)
+            return ov_client.certificate_rabbitmq.get_key_pair('default')
+        else:
+            logging.exception("Unexpected error")
+            raise e
+
+
+def generate_certificate_in_oneview(ov_client):
```
ricardogpsf: Is a private method, too? :grimacing:

oneview_redfish_toolkit/app.py
```diff
@@ -313,23 +313,7 @@ def hp_oneview_client_exception(exception):
...
-        if not scmb.is_cert_working_with_scmb():
-            logging.error('Failed to connect to scmb. Aborting...')
-            exit(1)
-        scmb_thread = threading.Thread(target=scmb.listen_scmb)
-        scmb_thread.daemon = True
-        scmb_thread.start()
-    else:
-        logging.warning("Authentication mode set to session. SCMB events will "
-                        "be disabled")
+    scmb.init_event_service()
```
ricardogpsf: Very good! :clap:
This turns clean this place of code and keeps the responsabilty of scmb logic with the scmb.py file. Nice!

oneview_redfish_toolkit/api/scmb.py
adriane-cardozo: I would also change to "RabbitMQ certificate and key pair retrieved from OneView."

PR #474, opened 2018-10-04
Improve error message for errors 500

karolcatunda:
- Improving 500 error message by including error description on generic message

### DIFF
--- a/oneview_redfish_toolkit/blueprints/util/response_builder.py
+++ b/oneview_redfish_toolkit/blueprints/util/response_builder.py
@@ -77,10 +77,12 @@ class ResponseBuilder(object):
 
     @staticmethod
     def error_500(error):
-        redfish_error = RedfishError(
-            "InternalError",
-            "The request failed due to an internal service error.  "
-            "The service is still operational.")
+        if hasattr(error, "description"):
+            msg = error.description
+        else:
+            msg = str(error)
+
+        redfish_error = RedfishError("InternalError", msg)
         redfish_error.add_extended_info("InternalError")
         return ResponseBuilder.response(redfish_error,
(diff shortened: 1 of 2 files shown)


oneview_redfish_toolkit/blueprints/util/response_builder.py
```diff
@@ -77,10 +77,12 @@ def error_404(error):
...
     @staticmethod
     def error_500(error):
-        redfish_error = RedfishError(
-            "InternalError",
-            "The request failed due to an internal service error.  "
-            "The service is still operational.")
+        if hasattr(error, "description"):
+            msg = error.description
+        else:
+            msg = str(error)
```
victorhugorodrigues: Can we create unit tests raising both types of error?

karolcatunda: Yes, we can. I will create unit tests for response builder file.

oneview_redfish_toolkit/tests/blueprints/test_manager_collection.py
```diff
@@ -43,19 +44,23 @@ def setUpClass(self):
...
+            Exception("An exception was occurred")
 
         with open(
                 'oneview_redfish_toolkit/mockups/errors/'
                 'Error500.json'
         ) as f:
             error_500 = json.load(f)
 
+        error_500_excep = copy.deepcopy(error_500)
+        error_500_excep["error"]["message"] = "An exception was occurred"
```
ricardogpsf: My english is not so good, but the message should be *"An exception ~was~ has occurred"*, not?

karolcatunda: Yes, that is correct! I will fix it. Thanks 😄

PR #475, opened 2018-10-04
Improvement performance for multiple OneViews

### LINKED ISSUE #464: Improvement performance for multiple OneViews
On toolkit request for a single resource, not collection, after check which OneView client use for the first OneView resource (serverHardware, serverProfileTemplate, etc.) we can use this same OneView client for all subsequent calls on OneView on the same request context.

victorhugorodrigues:
Closes #464 
Implementing Single OneView Context.
With this implementation the solution does not have to search in all OneViews after get the OneView that it's working with inside the same request context. Before this implementation every "get_all()" call to retrieve OneView data was requested on each one of multiple OneViews. With this implementation, after the application find any resource on one of multiple OneViews, all the subsequent calls will be requested only on that same OneView.

On the graph below there is the comparison before and after this implementation, for an Storage ResourceBlock request on toolkit API using three mulitple OneViews, reducing from **21** OneView requests needed to **13** requests, and elapsed time from **2.18** seconds to **0.88** seconds:

singleoneviewcontext

### DIFF
--- a/oneview_redfish_toolkit/multiple_oneview.py
+++ b/oneview_redfish_toolkit/multiple_oneview.py
@@ -78,22 +78,35 @@ def query_ov_client_by_resource(resource_id, resource, function,
         Returns:
             dict: OneView resource
     """
+    # Get OneView's IP in the single OneView context
+    single_oneview_ip = _is_single_oneview_context() and \
+        _get_single_oneview_ip()
+    # Get OneView's IP for cached resource ID
+    cached_oneview_ip = get_ov_ip_by_resource(resource_id)
+
     # Get OneView's IP in the single OneView context or cached by resource ID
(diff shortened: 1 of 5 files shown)


oneview_redfish_toolkit/multiple_oneview.py
```diff
@@ -176,8 +218,8 @@ def execute_query_ov_client(ov_client, resource, function, *args, **kwargs):
         try:
             result = ov_function(*args, **kwargs)
             return result
-        except Exception as e:
-            raise e
+        except Exception as exception:
```
ricardogpsf: We can remove the `except` clause, we are just raising the exception catched.
The code at the `finally` clause will always be executed even without the `except` clause.

PR #476, opened 2018-10-04
Abstracting map clients dict to a class

### LINKED ISSUE #444: Change the `map_clients` dict from `client_session.py` to be a class
We have to create a class to replace the [map_clients](https://github.com/HewlettPackard/oneview-redfish-toolkit/blob/master/oneview_redfish_toolkit/client_session.py#L50-L53) dict, its structure it's becoming complicated to understand: RedfishToken->OneViewIP->(OneviewClient, SessionID).

karolcatunda:
Closes #444

### DIFF
--- a/oneview_redfish_toolkit/client_session.py
+++ b/oneview_redfish_toolkit/client_session.py
@@ -44,6 +44,42 @@ GC_FREQUENCY_IN_SEC = 12 * 60 * 60
 lock = Lock()
 
 
+class MapClientsDict(object):
+
+    def set_ov_client_session(self, client_ov_by_ip):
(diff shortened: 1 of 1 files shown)


ricardogpsf:
IMHO, we can keep the code as before. The intention is have less complexity, but the new class just added more 5 methods and is a wrapper to same data structure that before, in other words, it has growed up the complexity. 😓 😟

victorhugorodrigues:
I think we can close this PR and we can further discuss how to improve this.

adriane-cardozo:
Closing, since we decided to keep the current code.

PR #477, opened 2018-10-08
Enhacement/update cache entry when item not found

### LINKED ISSUE #465: Update cache entry when resource moved to another OneView
for example: if a server hardware already mapped to Oneview 1 has moved to Oneview 2, future calls will always return 404, in this case the cache entry must be updated to OneView 2.

ricardogpsf:
Closes #465 

See the task to best description about the scenario.

### DIFF
--- a/oneview_redfish_toolkit/multiple_oneview.py
+++ b/oneview_redfish_toolkit/multiple_oneview.py
@@ -64,6 +64,12 @@ def set_map_resources_entry(resource_id, ip_oneview):
         get_map_resources()[resource_id] = ip_oneview
 
 
+def cleanup_map_resources_entry(resource_id):
+    with lock:
+        if resource_id in get_map_resources().keys():
+            del get_map_resources()[resource_id]
+
+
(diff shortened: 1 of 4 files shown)


oneview_redfish_toolkit/multiple_oneview.py
```diff
@@ -137,6 +158,8 @@ def search_resource_multiple_ov(resource, function, resource_id,
             resource: resource type (server_hardware)
             function: resource function name (get_all)
             resource_id: set only if it should look for a specific resource ID
+            ov_ips: List of Oneview IPs to search the resource. If None is
+                passed, the the list of IPs will be all IPs configured
```
adriane-cardozo: ov_ips: List of Oneview IPs to search for the resource. If None is passed, it will search in all IPs based on the toolkit configuration.

ricardogpsf:
@karolcatunda I will push the changes when all of you review the code, because the change requested by Adriane is very simple, it is just update a doc string.

ricardogpsf:
@karolcatunda @adriane-cardozo  done! The change were made.

PR #481, opened 2018-10-09
Include OneviewRedfishError handle on flask

### LINKED ISSUE #342: Include OneviewRedfishError handle on flask
Handle OneviewRedfishError exception on flask (app.py)

karolcatunda:
- Replacing OneViewRedfishError usages to errors handles on app.py
 - Removing oneview redfish error classes
 - Fixing unit tests according with this new approach

Closes #342

### DIFF
--- a/oneview_redfish_toolkit/api/computer_system.py
+++ b/oneview_redfish_toolkit/api/computer_system.py
@@ -16,8 +16,9 @@
 
 import collections
 from copy import deepcopy
+from flask_api import status
+from werkzeug.exceptions import abort
 
-from oneview_redfish_toolkit.api.errors import OneViewRedfishError
 from oneview_redfish_toolkit.api.redfish_json_validator \
     import RedfishJsonValidator
 from oneview_redfish_toolkit.api.resource_block_collection import \
@@ -218,9 +219,9 @@ class ComputerSystem(RedfishJsonValidator):
             server_profile_template)
 
         if storage_blocks and not controller:
-            raise OneViewRedfishError(
-                "The Server Profile Template should have a valid "
-                "storage controller to use the Storage Resource Blocks passed")
+            abort(status.HTTP_500_INTERNAL_SERVER_ERROR,
+                  "The Server Profile Template should have a valid "
+                  "storage controller to use the Storage Resource Blocks passed")
 
         for index, storage_block in enumerate(storage_blocks):
             storage_id = index + 1
--- a/oneview_redfish_toolkit/api/computer_system.py
+++ b/oneview_redfish_toolkit/api/computer_system.py
@@ -16,9 +16,9 @@
 
 import collections
 from copy import deepcopy
-from flask_api import status
 
-from oneview_redfish_toolkit.api.errors import OneViewRedfishException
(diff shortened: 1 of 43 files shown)


oneview_redfish_toolkit/api/computer_system.py
```diff
@@ -16,8 +16,9 @@
 
 import collections
 from copy import deepcopy
+from flask_api import status
+from werkzeug.exceptions import abort
```
victorhugorodrigues: You can use: from flask import abort
You could use this same library on others changes as well (Just to not repeat the same comment)

victorhugorodrigues:
I think that api modules (api/*.py) should not be responsible for the endpoint request execution pipeline. That means the api modules should not be responsible for call an abort and stop the request execution, it should be just responsible for translate OneView data to Redfish data and doesn't have visibility on request life cycle.
So my suggestion is, keep raising an custom Exception(OneviewRedfishError or ApiRedfishError) on the api modules, and the blueprint modules could or not handle this exception for special treatments. And create a new exception handle for this exception on app.py, for the cases where blueprint doesn't need specifically handle an Api error.

ricardogpsf:
I agree totally with @victorhugorodrigues said!

oneview_redfish_toolkit/api/computer_system.py
```diff
@@ -218,9 +219,11 @@ def _build_sas_logical_jbods(server_profile_template, storage_blocks):
             server_profile_template)
 
         if storage_blocks and not controller:
-            raise OneViewRedfishError(
-                "The Server Profile Template should have a valid "
-                "storage controller to use the Storage Resource Blocks passed")
+            raise OneViewRedfishException(
+                "The Server Profile Template should have a valid storage "
+                "controller to use the Storage Resource Blocks passed",
+                status.HTTP_403_FORBIDDEN
```
ricardogpsf: I think the api should not know about the HTTP status. It should just raise a exception, with some message, but doesn't specifying some HTTP status, the blueprint is who should catch the exception and return to the client the HTTP status properly.

oneview_redfish_toolkit/api/network_device_function.py
```diff
@@ -59,8 +59,11 @@ def __init__(self, device_id, device_function_id, server_hardware):
             virtual_port = self.get_resource_by_id(
                 port["virtualPorts"], "portNumber", virtual_port_number)
         except Exception:
-            raise OneViewRedfishResourceNotFoundError(
```
ricardogpsf: In this case, I believe the `OneViewRedfishResourceNotFoundError` is a good exception to raise.

oneview_redfish_toolkit/api/errors.py
```diff
@@ -24,19 +24,8 @@
                        'Session.INVALID']
 
 
-class OneViewRedfishError(Exception):
+class OneViewRedfishException(Exception):
 
-    def __init__(self, msg):
+    def __init__(self, msg, status_code_error):
```
ricardogpsf: As I had mentioned before, I think the api should not know about the HTTP status. This is a information of the "HTTP layer" (blueprint).

PR #486, opened 2018-10-10
Extend additional check for storage controller settings

### LINKED ISSUE #478: Extend additional check for storage controller settings
Check if storage controller mode is "HBA" or "Mixed" to include Storage Resource Blocks in a Zone.
This change may affect the composition request.

Confirm: When creating a server profile with a template that has more than a SAS controller, get the first one.

karolcatunda:
- Adding filter for controller mode, checking if it is HBA or Mixed

Closes #478

### DIFF
--- a/oneview_redfish_toolkit/services/computer_system_service.py
+++ b/oneview_redfish_toolkit/services/computer_system_service.py
@@ -48,7 +48,11 @@ class ComputerSystemService(object):
                 server_profile_tmpl: the Server Profile Template
         """
         for controller in server_profile_tmpl["localStorage"]["controllers"]:
-            if controller["deviceSlot"] != "Embedded":
+            has_valid_controller_mode = \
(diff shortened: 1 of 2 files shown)


oneview_redfish_toolkit/services/computer_system_service.py
```diff
@@ -48,7 +48,11 @@ def get_storage_controller(server_profile_tmpl):
                 server_profile_tmpl: the Server Profile Template
         """
         for controller in server_profile_tmpl["localStorage"]["controllers"]:
-            if controller["deviceSlot"] != "Embedded":
+            has_valid_controller_mode = \
```
adriane-cardozo: This check should also be applied on composition request and building a Zone.

victorhugorodrigues:
Maybe we could we have some unit tests for this scenarios

ricardogpsf:
I miss test to cover the "Mixed" mode (or at least put the "Mixed" mode in the json mockups).

PR #515, opened 2018-10-31
Add RedfishVersion in the log files

### LINKED ISSUE #511: Add RedfishVersion in the log files
To be able to check which release is installed.

ricardogpsf:
Closes #511


oneview_redfish_toolkit/app.py
```diff
@@ -302,6 +303,8 @@ def oneview_redfish_exception(exception):
 
         return ResponseBuilder.oneview_redfish_exception(exception)
 
+    logging.info("RedfishVersion : " + oneview_redfish_toolkit.version())
```
adriane-cardozo: We should add RedfishVersion in all debug files.

ricardogpsf: Is necessary put it in all files? I mean, each logger is used to write different logs, according with its own context.
Maybe write using the root logger is enough. 
I think write a application information (like version) using all loggers is redundance. What do you think?

ricardogpsf: @adriane-cardozo @karolcatunda

adriane-cardozo: Ok. Got it! We just need to make sure all logs are sent for analysis when needed.

PR #528, opened 2018-12-07
Adding implementation of Volume end point - issue 479

shobhit-sinha:
Signed-off-by: shobhit-sinha <shobhit.sinha@hpe.com>


shobhit-sinha:
incorrect check-ins went into...so closing it and will submit a new one

PR #529, opened 2018-12-10
issue 479-Adding redfish implementation for volumes

shobhit-sinha:


oneview_redfish_toolkit/api/storage.py
```diff
@@ -128,6 +128,9 @@ def build_for_composed_system(server_profile, server_hardware_type,
             attrs["Drives"].append({
                 "@odata.id": attrs["@odata.id"] + "/Drives/" + drive_id
             })
+        if len(sas_logical_jbods) != 0:
```
adriane-cardozo: I thought "Volumes" should be added only if you have logical drives in the server profile. Not sure if it makes sense present Volumes when there are only sas logical JBODs. @bsoutham, @sidneifranco  what do you think?

oneview_redfish_toolkit/api/volume.py
```diff
@@ -0,0 +1,202 @@
...
+            flag = False
+            for drivebay in drive_enclosure_object["driveBays"]:
+                for drivepath in drivebay["drive"]["drivePaths"]:
+                    if drivepath == path:
+                        flag = True
+                        drivebayuri = drivebay["uri"]
+                        drivebayuris.append(drivebayuri)
+                        break
+                if flag:
+                    break
```
victorhugorodrigues: You can do this code block using just two **for** loops instead of three, by removing the more external **for** loop, and on the **if** at line 102, you could change it to "**if drivepath in drivepaths:**"
And you can extract this code block into a new method for remove the flags.

PR #539, opened 2019-01-28
Fix SCMB

shobhit-sinha:


Vijayakannan-M:
These changes are to enable persistent subscription model for red-fish-toolkit.

1. In conf mode it will publish the message to persisted subscriber when the application starts
2. In session mode when the login process is done using given credential it will publish the message on the persisted subscriber.

